PR #45, opened 2021-05-18
Bump Go to 1.16

tri-adam:
Use Go 1.16 in CI. Replace `syscall` package usage for `Setresuid()`/`Setresgid()` in test helpers with `unix` package. As of Go 1.16, the `syscall` functions now include wrappers that ensure all OSThreads mirror a common system call (related discussion [here](https://github.com/golang/go/issues/1435)). This is not compatible with the approach we have taken in our test cases. The `unix` package retains the original semantics, so we use that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ orbs:
 parameters:
   go-version:
     type: string
-    default: 1.15.12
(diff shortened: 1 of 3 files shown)


dtrudg (approved):
LGTM

PR #51, opened 2021-05-21
Use Latest Debian in CI

tri-adam:
Rename `check-stretch` to `check-debian`. Update `golang:stretch` to `golang:buster`, the most recent stable version of Debian.


dtrudg (approved):
LGTM... bullseye will be out soon.

PR #98, opened 2021-06-10
Fix TestSIFInitializer Unit Test

tri-adam:
Fix issue with `TestSIFInitializer` unit test where `img.Writable` was inconsistent with the `mode` passed to `os.OpenFile`, causing a false negative error for the `NoPartitionSIF` test.

Leverage subtests to improve readability of unit test results. Simplify error check logic.


dtrudg (approved):
:+1:

PR #99, opened 2021-06-10
Remove Darwin Support

### LINKED ISSUE #90: Remove Darwin Support
The SingularityCE code base currently has partial support on Darwin (macOS), aimed at supporting Singularity Desktop (Singularity on macOS via virtualization.) Although Singularity Desktop is no longer maintained, we continue to build, test and maintain Darwin code in the SingularityCE code base.

It is of course possible to run Singularity on Darwin via Vagrant, as documented in the [SingularityCE Admin Guide](https://sylabs.io/guides/3.8/admin-guide/installation.html#installation-on-windows-or-mac). A notable exception is Macs built on Apple Silicon (#66), but since Singularity Desktop has never been supported on this platform either, this is not a loss of support, rather a continued absence.

Given the above, I believe we should consider removing Darwin support from the code base.

tri-adam:
Fixes #90


dtrudg (approved):
LGTM and tests cleanly in a couple of VM environments with/without apparmor/selinux etc.

PR #100, opened 2021-06-10
Include config.h Directly in Go Source

tri-adam:
Rather than include `config.h` through `CGO_CFLAGS` via `mconfig`, include directly in relevant source via a `#cgo` directive. This allows Go tools that are not aware of Makeit to parse/compile the source.


dtrudg (approved):
LGTM

PR #136, opened 2021-07-02
Allow --writable-tmpfs for %test step of build

### LINKED ISSUE #109: --writable-tmpfs for %test in singularity build
Hi,

Some of the software I want to embed in my image writes systematically in `/var/log/$dedicated_directory` when starting, even if just to display the version as I use in the `%test` section of my recipe. It is not a problem when using the image itself with a `--writable` flag or similar, but I can't find a way to test my software is properly installed at build time as I always get errors related to attempts to write on a read-only file system.

The ability to use `--writable-tmpfs` in `singularity build` for the test section as available for `singularity shell` or `singularity exec` would solve my problem and allow more flexibility in image testing.

I also considered to use `--bind`, but the directory used by the problematic software writes to a subdirectory of `/var/log` which do not exist at the moment the binding is done, so my only option is to bind the whole `/var/log` which I feel to be inadequate. 

I am using singularity 3.8.0 with a docker debian:buster bootstrap.

Best regards,
Sylvain

dtrudg:

Some software will unavoidably write to the container fs during reasonable tests. Using a build time bind to permit this is problematic as it is host specific (path available to bind), and the files created by the test likely shouldn't really stick around. Allow specifying `--writable-tmpfs` for a `singularity build` to use a writeable tmpfs in the `%test` section of the build only.

 - Fixes #109

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # SingularityCE Changelog
 
+## Changes since last release
+
+### New features / functionalities
+
+  - `--writable-tmpfs` can be used with `singularity build` to run
+    the `%test` section of the build with a ephemeral tmpfs overlay,
+    permitting tests that write to the container filesystem.
+
 ## v3.8.0 [2021-05-26]
 
 This is the first release of SingularityCE 3.8.0, the Community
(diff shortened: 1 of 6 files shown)


dtrudg:
Code is working, but test failure is due to this not being supported with fakeroot... I need to catch that combination and error before the build starts. Will address next week.

tri-adam (approved):
LGTM

PR #144, opened 2021-07-06
[WIP] nvcli POC / workup

dtrudg:
POC / workup for use of nvidia-container-cli


tri-adam:
> The device binding / effect of `NVIDIA_VISIBLE_DEVICES` as above is confusing. What's the best way to reconcile it with Singularity's historic behavior?

Hrm, yeah, tricky one. I don't see a big issue when `--contain` is used and `NVIDIA_VISIBLE_DEVICES` is set, as the expected semantics are likely more inline with the new behaviour? The case where `--contain` is used without `NVIDIA_VISIBLE_DEVICES` being set is certainly a big change. I think at a minimum we'd need a warning to be displayed that directs users to set that env var. I suppose the other approach would be to force the old behaviour in this case. Are there any other options you can think of @dtrudg?

dtrudg:
> > The device binding / effect of `NVIDIA_VISIBLE_DEVICES` as above is confusing. What's the best way to reconcile it with Singularity's historic behavior?
> 
> Hrm, yeah, tricky one. I don't see a big issue when `--contain` is used and `NVIDIA_VISIBLE_DEVICES` is set, as the expected semantics are likely more inline with the new behaviour? The case where `--contain` is used without `NVIDIA_VISIBLE_DEVICES` being set is certainly a big change. I think at a minimum we'd need a warning to be displayed that directs users to set that env var. I suppose the other approach would be to force the old behaviour in this case.

As it is we have a warning on the `--contain` case if `NVIDIA_VISIBLE_DEVICES` is not set...

```
	// When we use --contain we don't mount the NV devices by default in the nvidia-container-cli flow,
	// they must be mounted via specifying with`NVIDIA_VISIBLE_DEVICES`. This differs from the legacy
	// flow which mounts all GPU devices, always.
	if (IsContained || IsContainAll) && os.Getenv("NVIDIA_VISIBLE_DEVICES") == "" {
		sylog.Warningf("When using nvidia-container-cli with --contain NVIDIA_VISIBLE_DEVICES must be set or no GPUs will be available in container.")
	}
```

> Are there any other options you can think of @dtrudg?

I'm not sure there's anything else that makes sense? It's going to be a bit weird due to our differing isolation model than Docker. Arguable we might want a way of respecting `NVIDIA_VISIBLE_DEVICES` without the other things inferred by `--contain` too. 

I think this is probably worth a mailing list / Slack appeal, and bringing up in a community meeting at some point.

PR #153, opened 2021-07-15
fix: Allow env vars with literal $ via \\\$ in SINGULARITYENV_

### LINKED ISSUE #43: Escaped strings being expanded in environment variables causing $LIB LD_PRELOAD isssue
**Version of Singularity** 3.7.3

**Describe the bug**
Unable to set ``LD_PRELOAD=/foo/bar/$LIB/baz.so`` in an arbitrary container.

``$LIB`` is not another environment variable in this context, but is intended to be resolved by the linker, ``ld.so``. When set, using ``export SINGULARITYENV_LD_PRELOAD='/foo/bar/$LIB/baz.so'``, ``$LIB`` is treated as environment variable (which is unlikely to be set) and the variable that is set in the container is ``LD_PRELOAD=/foo/bar//baz.so`` rather than the hoped for result. As a result the library is not loaded and an error is output ``ERROR: ld.so: object '/foo/bar//baz.so`' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.`

**To Reproduce**
The underlying cause can be reproduced with any environment variable with an escaped string containing ``$``, however the behaviour in many cases could well be desired/by design. I have attached some example code that builds an ``LD_PRELOAD`` library that outputs a string when loaded using the ``$LIB`` variable attached.

[preload_test.tar.gz](https://github.com/sylabs/singularity/files/6493230/preload_test.tar.gz)

It requires a singularity image to test - ``ubuntu.simg`` in the same directory.
If you run ``make singularity_test`` then it should build the library and will run a hello world program using ``SINGULARITYENV_LD_PRELOAD``, which will error before outputting ``Hello world!``. It will then re-run the binary using a script in the container that sets ``LD_PRELOAD`` that works as expected and outputs ``Loaded library successfully`` before the output of the Hello World program.

**Expected behavior**
There should be a way to escape ``$`` characters in environment variables so that they can be set in the container. Alternatively there could be an exception made for ``$LIB`` or ``${LIB}`` in ``LD_PRELOAD``.

**OS / Linux Distribution** Ubuntu 20.04.2 LTS

**Installation Method** Source (from release 3.7.3 on github)

**Additional context**
This used to work in v3.5.3 and I believe the behaviour was changed in v3.6. I mentioned this problem on the Slack channel and @dtrudg asked me to create an issue.

dtrudg:

From Singularity 3.6, SINGULARITYENV_ prefixed env vars on the host become env vars in the container by:

 - Creating an injectable script which will `export KEY=VAL`.
 - Sourcing this script in the container setup, through the embedded shell interpreter.

When the injectable script was written, the value for each export statement was being quoted automatically using the Go `q` Printf style formatting. This handles double quoting the value, but also escapes other things in the value per Go string syntax.

It would be expected that a `\\\$LIB` in a SINGULARITYENV_ export at the host shell would resolve to a literal `$LIB` in the container...

 - In the host shell `\\\$LIB` is evaluated and becomes `\$LIB`, which will be used as the VAL in the injectable script.
 - In the Singularity embedded shell `\$LIB` is then evaluated and becomes a literal `$LIB` in the container env var.

However, the use of `q` Go string formatting means that `\$LIB` is greedily escaped according to Go string syntax expectations to become `\\$LIB` in the injectable script. This results in the container env var being `\` plus the evaluated value of an env var `$LIB`, rather than a literal `$LIB`.

To allow literal `$` in a container env var we:

 - Manually double quote the VAL in the injectable script export statement.
 - Manually escape only embedded `"` in the env var value, so that the meaning of other shell escapes is not modified.

This allows:

```
$ export SINGULARITYENV_LD_PRELOAD="/foo/bar/\\\$LIB/baz.so"
$ singularity run library://alpine env | grep PRELOAD
LD_PRELOAD=/foo/bar/$LIB/baz.so
```

Or if you bypass the host shell interpretation via exporting the SINGULARITYENV_ with single quotes:

```
$ export SINGULARITYENV_LD_PRELOAD='/foo/bar/\$LIB/baz.so'
$ singularity run library://alpine env | grep PRELOAD
LD_PRELOAD=/foo/bar/$LIB/baz.so
```

 - Fixes #43

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@
     the `%test` section of the build with a ephemeral tmpfs overlay,
     permitting tests that write to the container filesystem.
 
+### Bug Fixes
+
+  - Allow escaped `\$` in a SINGULARITYENV_ var to set a literal `$` in
+    a container env var.
+
 ## v3.8.0 [2021-05-26]
 
 This is the first release of SingularityCE 3.8.0, the Community
--- a/e2e/env/env.go
+++ b/e2e/env/env.go
@@ -446,5 +446,6 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
 		"environment file":         c.singularityEnvFile,
 		"issue 5057":               c.issue5057, // https://github.com/sylabs/hpcng/issues/5057
 		"issue 5426":               c.issue5426, // https://github.com/sylabs/hpcng/issues/5426
+		"issue 43":                 c.issue43,   // https://github.com/sylabs/singularity/issues/43
 	}
 }
--- a/e2e/env/regressions.go
+++ b/e2e/env/regressions.go
@@ -99,3 +99,26 @@ func (c ctx) issue5057(t *testing.T) {
 		e2e.ExpectExit(0),
 	)
 }
+
+// https://github.com/sylabs/singularity/issues/43
+// If a $ in a SINGULARITYENV_ env var is escaped, it should become a
+// literal $ in the container env var.
+// This allows setting e.g. LD_PRELOAD=/foo/bar/$LIB/baz.so
+func (c ctx) issue43(t *testing.T) {
+	e2e.EnsureImage(t, c.env)
+
+	env := []string{`SINGULARITYENV_LD_PRELOAD=/foo/bar/\$LIB/baz.so`}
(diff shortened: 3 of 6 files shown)


tri-adam (approved):
Nice. Love the detailed explanation, and clean solution!

PR #244, opened 2021-08-12
fix: honor hostname in library:// URI for `singularity delete`

### LINKED ISSUE #243: delete does not honor hostname in library URI
**Version of Singularity**

3.8.1

**Describe the bug**

`singularity delete` will accept a library URI with a hostname, e.g. 

```
$ singularity -d delete library://library.example.com/dtrudg-sylabs2/test/test
```

However, the request is made against the current remote.

**Expected behavior**

Specified hostname should be honored.

dtrudg:
`singularity delete` would accept a hostname in the library:// URI
provided, but would perform the request against the default remote.

Fix this, so it is handled in the same way as for push/pull.

 - Fixes #243 

### DIFF
--- a/cmd/internal/cli/delete.go
+++ b/cmd/internal/cli/delete.go
@@ -97,6 +97,18 @@ var deleteImageCmd = &cobra.Command{
 			sylog.Fatalf("Error parsing library ref: %v", err)
 		}
 
+		if deleteLibraryURI != "" && imageRef.Host != "" {
+			sylog.Fatalf("Conflicting arguments; do not use --library with a library URI containing host name")
+		}
+
+		var libraryURI string
+		if deleteLibraryURI != "" {
+			libraryURI = deleteLibraryURI
+		} else if imageRef.Host != "" {
+			// override libraryURI if ref contains host name
+			libraryURI = "https://" + imageRef.Host
+		}
+
 		r := fmt.Sprintf("%s:%s", imageRef.Path, imageRef.Tags[0])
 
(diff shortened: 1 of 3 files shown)


cmd/internal/cli/delete.go
```diff
@@ -97,6 +97,14 @@ var deleteImageCmd = &cobra.Command{
 			sylog.Fatalf("Error parsing library ref: %v", err)
 		}
 
+		var libraryURI string
+		if deleteLibraryURI != "" {
+			libraryURI = deleteLibraryURI
+		} else if imageRef.Host != "" {
+			// override libraryURI if ref contains host name
+			libraryURI = "https://" + imageRef.Host
```
tri-adam, line 105: I'm not sure what should happen if someone supplies `--library` as well as `library://host/blah`. I'm tempted to say we might want to treat this as an error? If the purpose of the `imageRef` is to (possibly) specify its location, it's confusing if it _doesn't_ delete from the right location because `--library` is set?

tri-adam (approved):
Looks good, just one question on an edge case, thanks.

cmd/internal/cli/delete.go
dtrudg: Yeah - this is a silly oversight on my part as it's already handled that way for `pull`

dtrudg: Please see amended code.

tri-adam (approved):

PR #527, opened 2022-01-25
fix: Don't prompt y/n to overwrite output in noninteractive build

### LINKED ISSUE #521: Opposite of flag to "--force". IE, for headless builds, a flag to die on any error instead of prompting for user input.

I wish to build my images inside a batch job on a cluster. If I start the batch job and tell it to build an image file that already exists, it tries to prompt for user input and then hangs indefinitely.

I'd like an additional commanline parameter such as "--noninteractive" or "--dieonerror" or whathaveyou.

I can wrap my builds in a shell script and make the shell script check for any situation that would cause `singularity build` to hang indefinitely.

**Additional context**

dtrudg:

If a `singularity build` is run without a terminal Stdin (i.e. non-interactive), fail quickly if the output file already exists. Don't get stuck trying to ask for a y/n that won't come.

This is a quick fix addressing this specific common location where a non-interactive build can get hung up... and there is a `--force` flag to be employed so nobody needs to pipe in a 'y'. 

Really there should be a proper review of where the interactive package is used, and sensible behavior of the functions for these types of uses. We maybe should have variant functions that e.g. fail or fall to a default if Stdin isn't a terminal. However we also need to consider usage that pipes in to answer a prompt (e.g. provide encryption passphrase) from another tool.

Will open a tech-debt issue for the above.

 - Fixes #521 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
   return code with `errnoRet` and `defaultErrnoRet`. Previously EPERM was hard
   coded. The example `etc/seccomp-profiles/default.json` has been updated.
 
+### Bug fixes
+
+- Don't prompt for y/n to overwrite an existing file when build is
+  called from a non-interactive environment. Fail with an error.
+
 ## v3.9.4 \[2022-01-19\]
 
 ### Bug fixes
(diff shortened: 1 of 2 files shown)


tri-adam (approved):
LGTM

PR #540, opened 2022-02-01
feat: Enable systemd cgroups management

### LINKED ISSUE #299: Systemd as cgroups manager

Follow up after #60 / #292 

systemd is the default init for all of our target distributions, and with cgroups v2 it would be useful to use it for cgroups management. This would allow cgroups applied to containers to fit neatly into the user slice / session scopes in the systemd managed v2 hierarchy.

Note that users can call `systemd-run ... singularity ...` to achieve something similar, and resource management is usually via a scheduler, so this may not be a high priority.

dtrudg:
Enable cgroups management via `systemd`.

`--apply-cgroups` for `shell/exec/run` etc. will create a `singularity-<pid>.scope` in the `system.slice`.

The OCI runtime will create a `singularity-oci-<name>.scope` in the `system.slice`.

This is a stepping stone toward allowing unpriv cgroups limits, via asking systemd to create and delegate scope in the user slice, on cgroups v2 systems.

Set systemd cgroups management as the default. It will be needed as we move toward rootless, and we are not directly supporting any non-systemd distributions at this time. It can be disabled, to fall back to direct manipulation of cgroupfs, by setting `systemd cgroups = no` in `singularity.conf`.

*NOTE...*

There is reasonable unit test coverage of the cgroups functionality with both cgroupfs and systemd management.

There is currently *NO* e2e coverage of systemd cgroups management. Our e2e tests are run in a new PID namespace. We can't ask the host systemd to manage cgroups for PIDs in this namespace, as they are completely separate from the host PIDs that system sees.

I am going to capture a tech debt issue for the e2e to address separately, as it won't be a trivial change (#563).

Fixes #299


internal/pkg/cgroups/util.go
```diff
@@ -26,6 +26,8 @@ func pidToPath(pid int) (path string, err error) {
 		return "", fmt.Errorf("cannot read %s: %w", pidCGFile, err)
 	}
 
+	fmt.Print(paths)
+
```
tri-adam, line 30: Is this left in intentionally?

dtrudg: Nope... I'll take care of that. Thanks!

PR #605, opened 2022-02-25
Adopt Go 1.18 (master)

tri-adam:
Bump Go versions used in CI to 1.17 and 1.18. Remove deprecated build constraint format.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ orbs:
 parameters:
   go-version:
     type: string
-    default: '1.17.7'
+    default: '1.18'
 
 executors:
   node:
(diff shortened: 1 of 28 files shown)


dtrudg:
E2E tests with go 1.18 will fail with...

```
FATAL:   Plugin compile failed with error: while generating plugin manifest: while loading plugin /tmp/plugin-3139335449/src/plugin.so: plugin.Open("/tmp/plugin-3139335449/src/plugin"): plugin was built with a different version of package github.com/sylabs/singularity/internal/pkg/util/user
```

Confirmed no issue under Go 1.17.8.

Did a very quick look at 'strings' output for the plugin.so vs singularity under 1.17.8 and 1.18 and there isn't any gross difference in the diffs.

`internal/pkg/util/user` does contain CGO code... so am suspecting something related to that at present.

tri-adam:
I'll re-base this now that #678 has been merged.

dtrudg (approved):
LGTM

Issue #563, opened 2022-02-09
E2E PID namespace workaround required for systemd cgroups tests


dtrudg:
**Type of issue**

technical debt

**Description of issue**

In PR #540 management of cgroups via systemd is introduced.

The E2E test suite runs in a PID namespace, as a means to ensure we don't have processes hanging around on the machine, and PIDs for instances etc. are fully isolated.

If we ask systemd to create/manage a cgroup from inside a PID namespace this will fail. We are providing it with a PID from inside the namespace, but systemd can only sees and manage PIDs outside of the namespace.

Either...

* Run the relevant tests outside of the E2E PID namespace (consider any conflicts with host).
* Run a separate systemd instance inside the E2E PID namespace (here be dragons).

PR #622, opened 2022-03-04
e2e: non-pid namespace e2e-tests for OCI/CGROUPS

dtrudg:
The Singularity e2e-tests were previously all run in a mount and PID
namespace, to avoid polluting the host filesystem (critical), and
process tree (less critical). These namespaces are set up in some CGO
init code.

The use of the PID namespace prevents testing with systemd as the
cgroups manager, as systemd is aware of the host process tree, not the
one in the new e2e PID namespace. This is a major omission since #540
switched to using systemd for cgroups management by default.

This PR:

 * Modifies the e2e init CGO code, so that an env var
   `SINGULARITYE_E2E_NO_PID_NS` will cause it *not* to create a new PID
   namespace.

 * Modifies the Makefile so that the e2e suite is called twice, once
   with PID ns, once without.

 * Moves the INSTANCE cgroups test into a new package/topic
   e2e/cgroups, run in the e2e call without PID ns.

 * Moves the OCI tests into the e2e call without PID ns.

 * Modifies the CGROUPS and OCI tests so that they test with both
   systemd and cgroupfs management, using a convenience wrapper function.

Fixes: #563 

Note - this breaks the e2e CLI coverage, as it only supports
collecting / analyzing one e2e run... where we now have two. The CLI
coverage metrics are unused in practice, and flawed (don't consider
combinations of flags required or possible), so I'm going to propose
removing them in a separate issue.


tri-adam (approved):
Seems reasonable to me.

PR #711, opened 2022-04-12
feat: squashfuse unpriv SIF image mount (experimental)

dtrudg:
Allow SIF images to be mounted with `squashfuse` in the unpriv / user-namespace flow.

The approach take here is to:

* Leverage the `MountFUSE` / `UnmountFUSE` functionality from sylabs/sif, rather than duplicating code.
* Perform the mount onto a temporary directory in the CLI layer, at the point where the SIF would otherwise be extracted. The singularity runtime then sees a sandbox.
* Add a 'CLEANUP_HOST' process that is forked from `starter` early, before any namespace manipulation. This process is able to unmount the SIF at cleanup time. A socket from the host cleanup process to the master process is used for coordination.

Mount at the CLI layer, rather than inside of the runtime engine, is used so that we can adopt a similar approach across `singularity oci` commands, and general use of runc in future.

`--fakeroot` is not currently supported.

`instance start` is not currently supported.

Fixes https://github.com/sylabs/singularity/issues/718

**Unsolved questions?**

* Should we have a `singularity.conf` to always try? It  *is* experimental, and there's a flag + env-var... but other experimental features have had a conf option in the past.

```
$ singularity run -u --sif-fuse ubuntu_latest.sif 
INFO:    Mounting SIF with FUSE...
Singularity> echo "HELLO!"
HELLO!
Singularity> 
exit
INFO:    Unmounting SIF with FUSE...
INFO:    Removing image tempDir /tmp/rootfs-949762240
```

### DIFF
--- a/cmd/internal/cli/actions_linux.go
+++ b/cmd/internal/cli/actions_linux.go
@@ -6,6 +6,7 @@
 package cli
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -17,6 +18,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	sifuser "github.com/sylabs/sif/v2/pkg/user"
 	"github.com/sylabs/singularity/internal/pkg/buildcfg"
 	"github.com/sylabs/singularity/internal/pkg/image/unpacker"
 	"github.com/sylabs/singularity/internal/pkg/instance"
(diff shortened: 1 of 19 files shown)


cmd/internal/cli/actions_linux.go
```diff
@@ -46,76 +49,149 @@ import (
...
+	squashfusePath, err := bin.FindBin("squashfuse")
+	if err != nil {
+		return fmt.Errorf("squashfuse is required: %w", err)
+	}
+	if _, err := bin.FindBin("fusermount"); err != nil {
+		return fmt.Errorf("fusermount is required: %w", err)
 	}
 
-	return tempDir, imageDir, err
+	return sifuser.Mount(context.TODO(), img.Path, imageDir,
```
tri-adam, line 191: Would be good to wire this further up the chain, perhaps `cobraCmd.Context()` in `execStarter`?

tri-adam (approved):
LGTM, small nit regarding contexts. Nice work!

tri-adam (approved):

PR #784, opened 2022-05-06
Add/container stats

### LINKED ISSUE #191: feature request: instance run statistics
I want to be able to get the run statistics of a singularity instance the same way that one can with `docker stats`, especially for memory usage and cpu usage.
 
Possibly a few more fields in `singularity instance list` or a new command like `singularity stats` which would return the statistics of specified running instances.

I guess one could use an external monitoring tool to track the usage of each instance, but it would be much better for it to exist within singularity.

**Additional context**
I don't think this exists yet, as far as I can tell from https://singularityhub.github.io/singularity-cli/commands-instances

btw -- awesome project and documentation, love the work!

vsoch:

This is the start of work to add `singularity instance stats <instance>` to show a similar view to docker stats, but for a singularity instance! As a test example:

image

The only wonky one is the CPU - and this is because the runc cgroups manager doesn't have any representation of percentages - what seems to be done with docker/podman is that you have to record two points in time and then calculate the difference. We can probably do that, but I was too lazy for this first shot. So for now - we have nanoseconds! :laughing: 

To reproduce using this locally you can do:

```bash
$ touch null.toml
$ sudo singularity instance start --apply-cgroups null.toml docker://busybox test
```
And then  from your singularity built from the PR:

```bash
$ sudo ./singularity instance stats test 
```
```bash
INFO:    Stats for test instance of /home/vanessa/go/src/github.com/sylabs/singularity/busybox_latest.sif (PID=742982)
INSTANCE NAME    CPU USAGE      MEM USAGE / LIMIT      MEM %    BLOCK I/O      PIDS
test             38112275 ns    6.773MiB / 15.33GiB    0.04%    713KiB / 0B    7
```
Note that you can also add `--json` to get a large dump of data!
```bash
$ sudo ./singularity instance stats --json test 
```
I had added the `--user` flag but if it's the case that this _always_ needs to be run with sudo I'm not sure it makes sense? Let me know.

 - Fixes #191 

And I think I need to add tests for a subset of this - I will wait for some maintainer wisdom and guidance about what to add and where because I'm not as familiar with the test setup yet as I should be.


internal/app/singularity/instance_linux.go
```diff
@@ -122,19 +130,129 @@ func WriteInstancePidFile(name, pidFile string) error {
 	return nil
 }
 
+// instanceListOrError is a private function to retrieve named instances or fail
+// We wrap the error from instance.List to provide a more specific error message
+func instanceListOrError(name, instanceUser string) ([]*instance.File, error) {
```
dtrudg, line 135: I probably wouldn't pull this out in a function at this point, but it's just a preference thing. Some thoughts:

- The order of arguments should match `instance.List` i.e. `instanceUser, name` so it's not confusing.
- Use `%w` to wrap the error, rather than `%v`
- Make sure the comment specifically says `fail if there are no instances`

dtrudg (commented):
This looks like a great start for instance stats.

> I had added the `--user flag` but if it's the case that this always needs to be run with sudo I'm not sure it makes sense? Let me know.

If the `--user` flag works to let the root user look at stats of other user's instances, then that's fine.

Technically this command should be sort-of useful for non-root users on newer systems... on modern systems there are other external ways to get an instance into a specific cgroup, and everything is in a systemd managed cgroup, generally.

Not sure how far you want to go in a first PR? Feel free to open as many issues as you can think of for stuff along the lines of `% CPU` calculation, interactive monitoring with a loop that updates every n seconds, showing stats for more all instances.

dtrudg:
The other thing we'd need to merge this is some kind of e2e test.

E2E tests that use `--apply-cgroups` with instances can be found at: https://github.com/sylabs/singularity/blob/master/e2e/cgroups/cgroups.go#L46

I'd suggest that a reasonably simple initial test for now would be a new `TestInstanceStats` function that:

* Starts an instance, with `--apply-cgroups` applying a memory limit.
* Calls `instance stats` for the instance, and examines the output with `e2e.ExpectOutput(e2e.ContainMatch .... )` to check that it does contain the expected instance name and memory limit.
* Stops the instance.

vsoch:
Woot, tests pass! :green_apple: :green_book: :green_circle: :green_heart: :green_heart:

cmd/internal/cli/instance_stats_linux.go
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2022, Sylabs Inc. All rights reserved.
```
dtrudg, line 1: This should almost certainly be your name, or your employer's if it was done under a circumstance in which they claim ownership.

cmd/internal/cli/instance_stats_linux.go
```diff
@@ -0,0 +1,82 @@
...
+		// An instance name is required
+		if len(args) == 0 {
+			return errors.New("invalid command")
+		}
+
+		uid := os.Getuid()
+
+		// Root is required to look at stats for another user
+		if instanceStatsUser != "" && uid != 0 {
+			sylog.Fatalf("Only the root user can look at stats of a user instance")
```
dtrudg, line 70:
```suggestion
			sylog.Fatalf("Only the root user can look at stats of a different user's instance")
```

cmd/internal/cli/instance_stats_linux.go
```diff
@@ -0,0 +1,82 @@
...
+	Usage:        "output stats in json",
+}
+
+// singularity instance stats
+var instanceStatsCmd = &cobra.Command{
+	Args:                  cobra.RangeArgs(0, 1),
+	DisableFlagsInUseLine: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// An instance name is required
+		if len(args) == 0 {
```
dtrudg, line 62: For consistency, it would be better to have cobra do this check, by using `ExactArgs(1)` instead of RangeArgs, above.

The command might be extended in future to accept no args (e.g. for stats on all instances), but setting the cobra Args to the current situation keeps usage and error handling more consistent.

e2e/cgroups/cgroups.go
```diff
@@ -41,6 +41,78 @@ type ctx struct {
...
+
+		// We always expect stats output, not create
+		createExitFunc := []e2e.SingularityCmdResultOp{}
+		instanceName := randomName(t)
+
+		// Start the instance with cgroups for stats
+		createArgs := append(tt.createArgs, instanceName)
+		c.env.RunSingularity(
+			t,
+			e2e.AsSubtest(tt.name+"/start"),
```
dtrudg, line 78: Instead of putting `/` delimited subtests here, it's more common to wrap everything inside the loop in an `t.Run(` call with an anonymous function, so that the structure of the nested tests is more apparent.

i.e.

```
for _, tt := range tests {
    t.Run(tt.name, func(t *.testing.T) {
          ...
          c.env.RunSingularity(
              t,
              e2e.AsSubtest("start"),
              ....
    })
}

File: internal/app/singularity/instance_linux.go [hunk 6]
```diff
@@ -122,19 +130,127 @@ func WriteInstancePidFile(name, pidFile string) error {
...
+
+	// Grab our instance to interact with!
+	i := ii[0]
+	if !formatJSON {
+		sylog.Infof("Stats for %s instance of %s (PID=%d)\n", i.Name, i.Image, i.Pid)
+	}
+
+	// Cut out early if we do not have cgroups
+	if !i.Cgroup {
+		url := "https://sylabs.io/guides/latest/user-guide/cgroups.html"
```
dtrudg, line 198: Could you change this to point people at the "user guide instances section" or similar.

A direct URL is nice, but will become misleading at some point. Many people trail quite a long way behind the latest version of SingularityCE. When we make rootless cgroups for instances the default, and update the docs, etc.  then the information there would be misleading for older versions.

We would likely document `instance stats` in the instances section of the guide, including a summary of the cgroups requirement, and links to more detail. It's called `Running Services` in the contents, at present - but needs to be changed so it's more obviously the instances section.... `Running Services (Instances)` or similar.

dtrudg (changes requested):
Thanks @vsoch - this is an awesome initial implementation.

A few comments which are just focused on fitting in with patterns in the code we'd like to maintain / encourage (rather than issues).

Thanks much for persevering with the e2e work, and fitting it into the structure used for the other cgroups tests.

dtrudg (approved):

PR #875, opened 2022-06-16
start of work to add DOCKER_HOST environment

### LINKED ISSUE #805: Singularity not honoring DOCKER_HOST env var when building from local image
**Version of Singularity**
What version of Singularity are you using? Run:
singularity-ce version 3.9.9

**Describe the bug**
Attempting to build a SIF from locally cached Docker image fails unexpectedly, claiming incorrectly that Docker isn't running.

My `docker.sock` is in a non-standard place, allowing me to run from multiple WSL2 instances. This is handled by setting the DOCKER_HOST env var in each instance. 

**To Reproduce**
```
singularity build pfoam-of7.sif docker-daemon://pfoam:of7
INFO:    Starting build...
FATAL:   While performing build: conveyor failed to get: loading image from docker engine: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
env | grep DOCKER
DOCKER_HOST=unix:///mnt/wsl/docker/docker.sock
docker images
REPOSITORY    TAG       IMAGE ID       CREATED        SIZE
pfoam         of7       cbd4a9a2bbcf   2 hours ago    3.97GB
```

**Expected behavior**
Build the SIF.

If I create a symlink at /var/run/docker.sock to the actual sock file, it works. This shouldn't be necessary. IMO.

**OS / Linux Distribution**
Ubuntu 20.04.4 LTS (via WSL2)

**Installation Method**
Source

**Additional context**
None

### LINKED ISSUE #878: Add support for Singularity to honor DOCKER_ envars
Currently, we look for these environment variables with a `SINGULARITY_` prefix, but we should also honor those set for Docker. I've taken a first shot at this in #875 and am opening a corresponding issue for tracking purposes.

vsoch:

This seems to be the basic to add support for a custom docker host as an envar or command line option. Some questions I have:

1. How/where to test?
2. Should the default value be set to something instead of empty string? If we use the dockerclient library we can just set the default to `dockerclient.DockerDaemonHost` which will populate based on the host-type (here is unix https://github.com/moby/moby/blob/a6919e12b103aab6eb95a67450b2a487bfec1097/client/client_unix.go) - but I figured it would be OK to test with the default not set since the [struct would do the same](https://github.com/containers/image/blob/7b9e2add30c3851b9195e1b44b47e06e66abc7a5/types/types.go#L641) if we didn't define it!

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

 - Fixes #805 
 - Fixes #878 
 
Note that I decided to open these together because full support for `DOCKER_HOST` would not work without the second.


dtrudg:
Hi @vsoch - this is looking great so far!

First thing to note is that as-is, I think we are not yet honoring `DOCKER_HOST`, but `SINGULARITY_DOCKER_HOST`, since the env keys are all namespaced under `SINGULARITY_`. This is fine for now. But I do think we should have the ability to read `DOCKER_HOST`, `DOCKER_USERNAME`, `DOCKER_PASSWORD`, without the prefix, for convenience / compatibility reasons That can be a separate issue though, which needs tackling in `pkg/cmdline/flag.go` as a follow-up.

> 1. How/where to test?

My thought is that in e2e/docker/docker.go we could add a test that...

- Does a `require.Command(t, "docker")` (from github.com/sylabs/singularity/internal/pkg/test/tool/require) ... to only run iif `docker` is available on the machine running the test.
- Invokes `docker build` to build an empty scratch image in the docker daemon.
- `singularity pull docker-daemon:xxxxxx` to pull the image (no `SINGULARITY_DOCKER_HOST`) - suceeds
- `singularity pull docker-daemon:xxxxxx` to pull the image (bad `SINGULARITY_DOCKER_HOST`) - fails
- `singularity pull docker-daemon:xxxxxx`` to pull the image (good `SINGULARITY_DOCKER_HOST`) - succeeds

> 2. Should the default value be set to something instead of empty string? If we use the dockerclient library we can just set the default to `dockerclient.DockerDaemonHost` which will populate based on the host-type (here is unix https://github.com/moby/moby/blob/a6919e12b103aab6eb95a67450b2a487bfec1097/client/client_unix.go) - but I figured it would be OK to test with the default not set since the [struct would do the same](https://github.com/containers/image/blob/7b9e2add30c3851b9195e1b44b47e06e66abc7a5/types/types.go#L641) if we didn't define it!

The default should be an empty string, as we want to allow `containers/image` to do the handling of defaults, not override that.

vsoch:
> That can be a separate issue though, which needs tackling in pkg/cmdline/flag.go as a follow-up.

okay - I'm going to add it here because I think we need this functionality to be able to say that DOCKER_HOST is truly supported.

vsoch:
oops forgot to run `make test`

vsoch:
Huzzah!! Green!!! :green_heart:

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
   specified instance, running within a cgroup.
 - Add `--sparse` flag to `overlay create` command to allow generation of a
   sparse ext3 overlay image.
+- Support for `DOCKER_HOST` parsing when using `docker-daemon://`
```
dtrudg, line 24: Also mention "DOCKER_USERNAME` and `DOCKER_PASSWORD` now supported without the `SINGULARITY_` prefix.

cmd/internal/cli/singularity.go
```diff
@@ -124,34 +125,48 @@ var singVerboseFlag = cmdline.Flag{
...
-	Name:         "docker-login",
-	Usage:        "login to a Docker Repository interactively",
-	EnvKeys:      []string{"DOCKER_LOGIN"},
+	ID:            "dockerLoginFlag",
+	Value:         &dockerLogin,
+	DefaultValue:  false,
+	Name:          "docker-login",
+	Usage:         "login to a Docker Repository interactively",
+	EnvKeys:       []string{"DOCKER_LOGIN"},
+	WithoutPrefix: true,
```
dtrudg, line 158:
```suggestion
```

I don't think this is a generally used flag / env var. It's specific to Singularity vs a command approach `docker login` in other tools.

e2e/docker/docker.go
```diff
@@ -118,6 +120,118 @@ func (c ctx) testDockerPulls(t *testing.T) {
...
+			exit:       255,
+		},
+
+		// Set to default should succeed
+		// The default host varies based on OS, so we use dockerclient default
+		{
+			name:       "singularityDockerHostValid",
+			envarName:  "SINGULARITY_DOCKER_HOST",
+			envarValue: dockerclient.DefaultDockerHost,
+			exit:       255,
```
dtrudg, line 198: I would expect this to check `exit: 0` for success.

e2e/docker/docker.go
```diff
@@ -118,6 +120,118 @@ func (c ctx) testDockerPulls(t *testing.T) {
...
+		},
+		{
+			name:       "dockerHostValid",
+			envarName:  "DOCKER_HOST",
+			envarValue: dockerclient.DefaultDockerHost,
+			exit:       255,
```
dtrudg, line 204: I would expect this to check `exit: 0` for success.

e2e/docker/docker.go
```diff
@@ -118,6 +120,118 @@ func (c ctx) testDockerPulls(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+
+		// Export to environment
+		if tt.envarValue != "" {
+			err := os.Setenv(tt.envarName, tt.envarValue)
```
dtrudg, line 212: Don't export to the global environment, as it will stick around and could affect other things if not cleaned up carefully.

The `RunSingularity` function can take environment via an `e2e.WithEnv` call, that will be limited to our single run of singularity. See e.g. here:

https://github.com/sylabs/singularity/blob/f1477058a25cddedb0ca5b2377b6f9ef827865a7/e2e/run/run.go#L98

dtrudg (changes requested):
Thanks @vsoch - this is going to be great to get merged!

Left some comments on the code for minor things. Two larger issues:

1. The e2e test isn't running.Because the e2e suite uses a custom approach, it doesn't automatically run anything that begins with `Test` like a standard `go test` run. You need to add the test here:

https://github.com/sylabs/singularity/blob/b63671732717634875ebd3e0150508ac896950f1/e2e/docker/docker.go#L845

2. You have added the flag `dockerHost`, and added `DockerDaemonHost` to the bundle.Options struct. I can see where you read that bundle.Options value in conveyorPacker_oci.go, but I don't see where you set it from the flag value. Have I missed something?

internal/pkg/build/sources/conveyorPacker_oci.go
```diff
@@ -144,11 +144,13 @@ func (cp *OCIConveyorPacker) Get(ctx context.Context, b *sytypes.Bundle) (err er
 	cp.sysCtx = &types.SystemContext{
 		OCIInsecureSkipTLSVerify: cp.b.Opts.NoHTTPS,
 		DockerAuthConfig:         cp.b.Opts.DockerAuthConfig,
+		DockerDaemonHost:         cp.b.Opts.DockerDaemonHost,
```
vsoch, line 147: @dtrudg is it not enough to add it here?

e2e/docker/docker.go
vsoch: copy paste failure!

internal/pkg/build/sources/conveyorPacker_oci.go
vsoch: At least in early testing, when I changed this cp.sysCtx variable to something nonsensical it was enough for it to try using it (and tell me localhost:pancakes wasn't reachable)!

vsoch:
> You have added the flag dockerHost, and added DockerDaemonHost to the bundle.Options struct. I can see where you read that bundle.Options value in conveyorPacker_oci.go, but I don't see where you set it from the flag value. Have I missed something?

Is that needed? My impression was that the options showed up there. Could you show me where to do that?

internal/pkg/build/sources/conveyorPacker_oci.go
dtrudg: You need to set it into the build options struct in `build_linux.go` here, so that it is propagated from the flag in the CLI level, down to the build level. Note how `DockerAuthConfig` is set here:

https://github.com/sylabs/singularity/blob/702930d9c36c8ff80eabdf1481c5c03a462b1a70/cmd/internal/cli/build_linux.go#L383

dtrudg:
> > You have added the flag dockerHost, and added DockerDaemonHost to the bundle.Options struct. I can see where you read that bundle.Options value in conveyorPacker_oci.go, but I don't see where you set it from the flag value. Have I missed something?
> 
> Is that needed? My impression was that the options showed up there. Could you show me where to do that?

I replied inline, above. You have to set the value from the flag into the Opts struct in build_linux.go, which is then passed down into the build.

On the e2e-tests, they are failing now with:

```
=== CONT  TestE2E/PAR/DOCKER/docker_host
    docker.go:155: Unexpected error while running command.
        exit status 125
```

Not a lot to go on here. It might be worth using something like the below, to see all the StdOut &StdErr from `docker` to debug more easily:

```
cmd := exec.Command("docker", "build", dockerURI, "-noappend", "-all-root")
cmd.Dir = tmpPath
out, err = cmd.CombinedOutput()
if err != nil {
	t.Fatalf("Unexpected error while running command %s: %s", err, string(out))
}
```

vsoch:
oh this seems to work
```
dockerHost := cmd.Flags().Lookup("docker-host").Value.String()
```
my lord that is so ugly.

dtrudg (approved):
Thanks @vsoch, this is great. I will open an issue to tidy up the privilege requirement in the e2e tests.

PR #879, opened 2022-06-17
feat: non-root, non --fakeroot builds with proot

### LINKED ISSUE #880: proot based non-root / non --fakeroot builds

Non-root users cannot build containers from definition files without `--fakeroot`, which requires user namespaces and subuid/subgid mapping. 

The `proot` tool allows unprivileged `ptrace` based emulation of the root user. It should be possible to perform the `%post` and `%test` steps of a build using `proot`.

`proot` is easily available as a static binary at https://proot-me.github.io/

Apptainer implemented non-root builds using the `fakeroot` command, binding it into the container that is being built, from the host. This has limitations, most notably there needs to be glibc compatibility between host `fakeroot` and the container `glibc`. See also #870 for discussion of this - you cannot e.g. build an Ubuntu 20.04 container on a new Fedora.

Charliecloud implemented unprivileged builds using an unpriv usernamespace (without subuid/subgid mapping), and injection of `fakeroot` into the container that is being built (Dockerfile). This requires distribution specific injection code, and sometimes some workarounds for the initial packaging install. See https://github.com/hpc/charliecloud/pull/857 and subsequent PRs.

dtrudg:

Non-root users can now build from a definition file, on systems that do not support `--fakeroot`. This requires the statically built `proot` command (<https://proot-me.github.io/>) to be available on the user `PATH`. 

These builds:
  - Do not support `arch` / `debootstrap` / `yum` / `zypper` bootstraps. Use `localimage`, `library`, `oras`, or one of the docker/oci sources.
  - Do not support `%pre` and `%setup` sections.
  - Run the `%post` sections of a build in the container as an emulated root user.
  - Run the `%test` section of a build as the non-root user, like `singularity test`.
  - Are subject to any restrictions imposed in `singularity.conf`.
  - Incur a performance penalty due to `proot`'s `ptrace` based interception of syscalls.
  - May fail if the `%post` and `%test` scripts require privileged operations that `proot` cannot emulate.

 - Fixes #880

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,6 +104,12 @@ commands:
       - run:
           name: Install dependencies
           command: <<# parameters.sudo >>sudo <</ parameters.sudo >>apt-get -q install -y build-essential squashfs-tools libseccomp-dev libssl-dev uuid-dev cryptsetup-bin runc libglib2.0-dev squashfuse
+      - run:
+          name: Install proot
+          command: |-
+            <<# parameters.sudo >>sudo <</ parameters.sudo >>curl -L -o /usr/local/bin/proot https://proot.gitlab.io/proot/bin/proot
+            <<# parameters.sudo >>sudo <</ parameters.sudo >>chmod +x /usr/local/bin/proot
+
   configure-singularity:
     steps:
       - run:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,20 @@
 - Instances started by a non-root user can use `--apply-cgroups` to apply
   resource limits. Requires cgroups v2, and delegation configured via systemd.
 - Added EL9 package builds to CI for GitHub releases.
+- Non-root users can now build from a definition file, on systems that do not
+  support `--fakeroot`. This requires the statically built `proot` command
+  (<https://proot-me.github.io/>) to be available on the user `PATH`. These builds:
+  - Do not support `arch` / `debootstrap` / `yum` / `zypper` bootstraps. Use
+    `localimage`, `library`, `oras`, or one of the docker/oci sources.
+  - Do not support `%pre` and `%setup` sections.
+  - Run the `%post` sections of a build in the container as an emulated root
+    user.
+  - Run the `%test` section of a build as the non-root user, like `singularity
+    test`.
+  - Are subject to any restrictions imposed in `singularity.conf`.
+  - Incur a performance penalty due to `proot`'s `ptrace` based interception of syscalls.
+  - May fail if the `%post` script requires privileged operations that `proot`
+    cannot emulate.
 
 ### Bug Fixes
 
--- a/cmd/internal/cli/action_flags.go
+++ b/cmd/internal/cli/action_flags.go
@@ -36,6 +36,7 @@ var (
 	SingularityEnv     map[string]string
 	SingularityEnvFile string
 	NoMount            []string
+	Proot              string
 
 	IsBoot          bool
 	IsFakeroot      bool
@@ -788,6 +789,17 @@ var actionSIFFUSEFlag = cmdline.Flag{
 	EnvKeys:      []string{"SIF_FUSE"},
 }
 
+// --proot (hidden)
+var actionProotFlag = cmdline.Flag{
+	ID:           "actionProot",
+	Value:        &Proot,
+	DefaultValue: "",
+	Name:         "proot",
+	Usage:        "Bind proot from the host into /.singularity.d/libs",
+	EnvKeys:      []string{"PROOT"},
+	Hidden:       true,
+}
+
 func init() {
 	addCmdInit(func(cmdManager *cmdline.CommandManager) {
 		cmdManager.RegisterCmd(ExecCmd)
@@ -880,5 +892,6 @@ func init() {
 		cmdManager.RegisterFlagForCmd(&actionOomKillDisableFlag, actionsInstanceCmd...)
 		cmdManager.RegisterFlagForCmd(&actionPidsLimitFlag, actionsInstanceCmd...)
 		cmdManager.RegisterFlagForCmd(&actionSIFFUSEFlag, actionsCmd...)
+		cmdManager.RegisterFlagForCmd(&actionProotFlag, actionsCmd...)
 	})
 }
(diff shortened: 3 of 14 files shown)


DrDaveD:
I suggest noting that the available generic statically compiled x86_64 proot binary that can be downloaded from the proot home page does not work on CentOS7.  I was not even able to compile a static copy there due to missing static libtalloc and python3.6m libraries even in the -devel packages.  I suppose it could be done by compiling libtalloc and python also from source.

dtrudg:
> I suggest noting that the available generic statically compiled x86_64 proot binary that can be downloaded from the proot home page does not work on CentOS7. I was not even able to compile a static copy there due to missing static libtalloc and python3.6m libraries even in the -devel packages. I suppose it could be done by compiling libtalloc and python also from source.

Looks lilke there is an issue with seccomp backports in the later EL7 kernels. The static `proot` should work with `PROOT_NO_SECCOMP=1` set, which is something we can add a workaround for.

DrDaveD:
Good find!  That indeed makes it run on centos7.

However, when I try to use it to install a package like openssh unprivileged into a centos7 container (the test case from the [Charliecloud SC21 paper](https://dl.acm.org/doi/10.1145/3458817.3476187)) it fails for me in the same way as just using `unshare -rm`: it gets an error
```
error: unpacking of archive failed on file /usr/libexec/openssh/ssh-keysign;62b45471: cpio: chown
```
which fakeroot avoids.

dtrudg:
Following up - here is a demo combining this `proot` approach, with an explicit `fakeroot` in `%post` to acheive the test case from the paper (centos 7, install openssh).

```
Bootstrap: docker
From: centos:7

%post
        yum -y install epel-release
        yum -y install fakeroot
        fakeroot yum -y install openssh
```

```
$ singularity build ssh.sif ssh.def
INFO:    Using proot to build unprivileged. Not all builds are supported. If build fails, use --remote or --fakeroot.
WARNING: 'nodev' mount option set on /tmp, it could be a source of failure during build process
INFO:    Starting build...

...snipped...

Installed:
  openssh.x86_64 0:7.4p1-22.el7_9                                                                                                  

Dependency Installed:
  fipscheck.x86_64 0:1.4.1-6.el7                                 fipscheck-lib.x86_64 0:1.4.1-6.el7                                

Complete!
INFO:    Creating SIF file...
INFO:    Build complete: ssh.sif
```

This is on a Fedora 36 host, where I can't achieve the build using the `fakeroot` bind approach:

```
Bootstrap: docker
From: centos:7

%post
        yum -y install openssh
```

```
$ apptainer build openssh-centos7.sif ssh.def 
WARNING: /usr/local/etc/singularity/ exists, migration to apptainer by system administrator is not complete
WARNING: /usr/local/etc/singularity/ exists, migration to apptainer by system administrator is not complete
WARNING: 'nodev' mount option set on /tmp, it could be a source of failure during build process
INFO:    Starting build...
Getting image source signatures
Copying blob 2d473b07cdd5 done  
Copying config 6d42e44a9f done  
Writing manifest to image destination
Storing signatures
2022/06/23 09:20:26  info unpack layer: sha256:2d473b07cdd5f0912cd6f1a703352c82b512407db6b05b43f2553732b55df3bc
2022/06/23 09:20:26  warn xattr{run/lock/lockdev} ignoring ENOTSUP on setxattr "user.rootlesscontainers"
2022/06/23 09:20:26  warn xattr{/tmp/build-temp-21654346/rootfs/run/lock/lockdev} destination filesystem does not support xattrs, further warnings will be suppressed
INFO:    Running post scriptlet
/usr/bin/faked-tcp: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /usr/bin/faked-tcp)
/usr/bin/faked-tcp: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /usr/bin/faked-tcp)
fakeroot: error while starting the `faked' daemon.
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
FATAL:   While performing build: while running engine: exit status 1
```

DrDaveD:
Ok but then is there anything that proot gives you that isn't available with `unshare -rm`?  Package installs like openssh was the whole reason for why I introduced invoking fakeroot in apptainer.  I guess if you don't want to depend on unprivileged user namespaces proot would be an option to that.  There's so many options & tradeoffs, it's hard to decide what to offer.

dtrudg:
> I guess if you don't want to depend on unprivileged user namespaces proot would be an option to that. 

This, exactly. We have some users who would like these kind of builds, but who do not enable unprivileged user namespaces, as a specific choice. There is a lot you _can_ install successfully without needing `fakeroot` on top of `proot`, and you can use both if you need to.

There have been some serious CVEs in the kernel related to unprivileged user namespaces, so some sites are reluctant to turn them on. They may take the position that it's much easier to quickly remove or patch a setuid binary for a container runtime that has an issues, than it is to update the kernel across a system. You can definitely argue the other way also, but we want to accommodate people as best we can on this.

DrDaveD:
For other people unprivileged user namespaces will probably be more convenient to use than proot, so it's probably good to provide both options.

Regarding preferring setuid to unprivileged user namespaces because of fears of CVEs,  I always recommend that people disable network namespaces if they can, because all the CVEs related to unprivileged user namespaces in the last few years have been from the combination of the two.

dtrudg:
> Regarding preferring setuid to unprivileged user namespaces because of fears of CVEs, I always recommend that people disable network namespaces if they can, because all the CVEs related to unprivileged user namespaces in the last few years have been from the combination of the two.

This is not the case. There have notably been a filesystem related, and a cgroups related, vulnerability this year. Good write-ups here:

* https://sysdig.com/blog/cve-2022-0185-container-escape/
* https://sysdig.com/blog/detecting-mitigating-cve-2022-0492-sysdig/

DrDaveD:
Ok.  I forgot about the first one because it didn't affect most of our users (most are using el7) and hadn't heard about the second one.

Ok so I'll change "all" in my statement to "most".

DrDaveD:
For the record, CVE-2022-0492 is not exploitable on Red Hat systems via unprivileged user namespaces as described in that second write-up, I tested it.  Red Hat's advisory does say it is applicable to 7 & 8 but doesn't mention namespaces, and says the attack complexity is high.  This is why the security teams I work with did not raise awareness of this vulnerability and why I hadn't noticed it.

EmmEff (approved):

PR #889, opened 2022-06-22
start of work to add singularity build from Dockerfile support!

### LINKED ISSUE #84: Build From Dockerfile
Some projects don't have ready built or up-to-date containers, but do have a Dockerfile. At present you cannot build these containers with Singularity without translating the Dockerfile to a Singularity definition file.

What about Singularity being able to build from a Dockerfile (with a limited subset of functionality for anything in a Singularity recipe that doesn’t map to a Docker directive) so the user doesn’t need to re-write recipes?

**Additional context**
Proposed on roadmap doc by @vsoch

vsoch:

This is a prototype or proof of concept that we can build from Dockerfile. The work here currently includes support for FROM/ADD/COPY/ENV/LABEL/CMD/ENTRYPOINT and is setup to also include support for multistage builds (although this needs to be finished and I need to look into how singularity handles this with respect to copying files between stages. There is no expectation of merging this, it was a lot of fun to work on / figure out and I learned a lot!

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

 - Fixes #84


vsoch:
> With respect to what's supported, COPY and ADD aren't quite the same. ADD can do files (like COPY), but also two things that Singularity's %file section won't:

Yep this I'm aware (I'm not a container noob @dtrudg!) I have a Singularity <-> Docker converter that I made years ago and for the time being, we treat COPY and ADD as the same, assuming that users don't know this distinction and use them to add files to the container (not with the special listing of actions). If someone goes outside that use case I consider it not supported as Singularity doesn't support it.

PR #964, opened 2022-09-07
Replace Deprecated ioutil Package

tri-adam:
Replace use of the `io/ioutil` package, which was deprecated in Go 1.16 ([ref](https://go.dev/doc/go1.16#ioutil)).


dtrudg:
LGTM and ready to merge. Will let you mark as RFR though @tri-adam, in case anything left?

tri-adam:
> LGTM and ready to merge. Will let you mark as RFR though @tri-adam, in case anything left?

Will re-base to resolve conflicts, and then should be good to go. Thank you!

PR #967, opened 2022-09-08
Replace Deprecated Title Function (release-3.10)

tri-adam:
Replace use of [func strings.Title](https://pkg.go.dev/strings#Title), which was deprecated in Go 1.18 ([ref](https://go.dev/doc/go1.18#strings)).


dtrudg:
Per https://github.com/containers/podman/pull/13999 - using the language packages adds quite a bit of binary size, for something that is pretty trivial here. Is there a naive way we can do this?

dtrudg:
Ignore my previous comment .... we definitely want to retain the case style of the user agent, and it's not un-trivial to modifyto a naive way catering to how it's different in PRO etc.

dtrudg (approved):

dtrudg (approved):

PR #969, opened 2022-09-08
E2E Test Timeouts

tri-adam:
Troubleshooting timeouts in `e2e-tests` job...


dtrudg:
Hopefully alleviated by #976 (pending)

dtrudg:
So far so good.... going to close this for now, as I don't think we need to bump the timeout or instance size. Reopen if I'm wrong @tri-adam :-)

PR #1013, opened 2022-10-05
refactor: move execStarter functionality out of CLI layer

dtrudg:

This PR removes the execStarter code from the CLI layer. execStarter was responsible for assembling CLI flags/options into an engine configuration, and calling the Singularity runtime's starter executable to launch the container with that configuration.

execStarter has been replaced with a Launcher struct, which is passed configuration via functional options, and implements an Exec function that performs container launch.

The intention of this PR is to remove the direct coupling of CLI and starter launch, so that next we can add a second OCI runtime launch method, with the CLI calling either the Singularity starter, or the OCI launch as appropriate.

The functional options / Launcher configuration are currently a naive translation of the previous direct use of CLI flag variables in execStarter. The launch package is currently in internal/pkg. The intention is that the interface will develop as the OCI launch functionality is written. Eventually there will be a stable interface for container launch, which can be exposed in pkg, to allow code other than the Singularity CLI to call it to launch containers.

Additionally, the PR makes the action flag variables private, and passes down a context into shell evaluation functions, where the other changes triggered a linter rule indicating this should be done.

### DIFF
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -21,6 +21,7 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/client/oci"
 	"github.com/sylabs/singularity/internal/pkg/client/oras"
 	"github.com/sylabs/singularity/internal/pkg/client/shub"
+	"github.com/sylabs/singularity/internal/pkg/runtime/launch"
 	"github.com/sylabs/singularity/internal/pkg/util/uri"
 	"github.com/sylabs/singularity/pkg/sylog"
 )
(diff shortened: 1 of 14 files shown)


tri-adam (approved):

cmd/internal/cli/actions.go
```diff
@@ -244,15 +251,90 @@ var TestCmd = &cobra.Command{
...
+	if err != nil {
+		return err
+	}
+
+	ki, err := getEncryptionMaterial(cmd)
+	if err != nil {
+		return err
+	}
+
+	opts := []launch.Option{
```
DrDaveD, line 288: I'm concerned that there will be future programming error cases where people forget to copy the cobra variable to the `launch.Option` structure when adding a new option.  Wouldn't it be possible/better to have cobra write directly into the launch structure?

Issue #1095, opened 2022-10-28
 Support sign/verify with X.509 certificates


donpellegrino:
**Description**

Singularity currently supports PGP/GPG web of trust for digital signatures. It would be useful to also support hierarchical chain of trust with X.509 certificates. This would provide advantages for users and system administrators having Certificate Authority (CA) certificates already in place in the OS, since they could trust signed containers without additional steps for verification of the signer.

**Desired Solution**
The ability to sign containers with X.509 certificates. The ability to verify containers against CA certificates already in place in the host OS.

**Describe alternatives you've considered**
PGP/GPG key management servers can provide another mechanism for trusting certificates. However, some organizations already have established CA cert. processes and lack PGP/GPG key management servers. It would be useful to support the organizations existing CA certs.

**Additional context**
Certificate providers such as IdenTrust provide code signing certificates. It would be useful to be consistent with the processes used for code signing and verification established by such providers and implemented in other software execution environments.

DrDaveD:
FYI there is currently a [pull request](https://github.com/apptainer/apptainer/pull/747) to do this pending on the Apptainer project.

tri-adam:
@donpellegrino, thank you for the request. Makes sense to me.

@DrDaveD, thanks for the pointer. That PR appears that depends on code not present in the [integrity package](https://pkg.go.dev/github.com/sylabs/sif/v2/pkg/integrity) used by Singularity. Not necessarily a blocker, but noting that here since it's relevant to the scope of work.

DrDaveD:
Right, that came in [another PR](https://github.com/apptainer/sif/pull/147) and a [followup PR](https://github.com/apptainer/sif/pull/150) in the apptainer fork of sif.

tri-adam:
Some notes on progress here:

- An on-disk format (DSSE) has been proposed for this work in a [CE Slack thread](https://singularityce.slack.com/archives/C020JBN4UDU/p1667325010847079) and a [CE Google Groups thread](https://groups.google.com/g/singularity-ce/c/HaX58-yAJ5M). Thanks to @fnikolai and @jscook2345 for contributing to the discussion.
- Support for DSSE encoding/decoding was merged in SIF: https://github.com/sylabs/sif/pull/228
- A public API for non-PGP signatures has been proposed in SIF: https://github.com/sylabs/sif/pull/245
- A draft PR is up with user-facing support for basic `sign`/`verify` using non-PGP public/private keys: #1120

[...]

PR #1120, opened 2022-11-04
Sign/Verify with Non-PGP Key Material

tri-adam:

Add `--key` option to `singularity sign` and `singularity verify`, which expect PEM-encoded key material that will be used to sign/verify the image. Add end-to-end tests to cover `--key` usage.

Extend `internal/app/singularity` package to support signing with a `signature.Signer` and verifying with a `signature.Verifier`.

Centralize test keys and images that are used by multiple packages to `test/`.

 - Related to #1095


cmd/internal/cli/sign.go
```diff
@@ -61,14 +65,23 @@ var signSifDescIDFlag = cmdline.Flag{
...
 }
 
+// --key
+var signPrivateKeyFlag = cmdline.Flag{
+	ID:           "privateKeyFlag",
+	Value:        &priKeyPath,
+	DefaultValue: "",
+	Name:         "key",
+	Usage:        "path to the private key file",
+}
```
dtrudg, line 75: Do we want to associate this with an env var also?

Potentially a user could set an env var to have the key used always, without needing to specify the flag, in an environment that requires that?

cmd/internal/cli/verify.go
```diff
@@ -78,6 +81,15 @@ var verifySifDescIDFlag = cmdline.Flag{
...
 }
 
+// --key
+var verifyPublicKeyFlag = cmdline.Flag{
+	ID:           "publicKeyFlag",
+	Value:        &pubKeyPath,
+	DefaultValue: "",
+	Name:         "key",
+	Usage:        "path to the public key file",
+}
```
dtrudg, line 91: Ditto the signing - do we want to expose with an env var also?

dtrudg (commented):
This is looking good to me. I have a couple of questions...

1. Should the '--key' flag have an associated env var, so that it's straightforward to set a default key for all verification operations etc.?

2. If the top level `test` directory is only going to contain test data, and not code for tests, would it be reasonable to use `testdata` instead? As I understand it, the go toolchain would then ignore any `.go` files in there.... which would make it a good place to 'hide' deprecated plugin examples that we may still need to test compile for a time, but don't want to keepin in `examples`, have appearing in go doc etc.

>Directory and file names that begin with "." or "_" are ignored by the go tool, as are directories named "testdata".

:+1: on centralising the test data. I think there might be some more that is essentially duplicated between unit and e2e tests, that should be found and dug out at some point when we have some time.

tri-adam:
> 1. Should the '--key' flag have an associated env var, so that it's straightforward to set a default key for all verification operations etc.?

Good idea, although this may introduce some subtlety for signing, since the type of signature being generated would then depend on the environment. Perhaps it would be worth adding some visual feedback?

Environment variable not set:

```sh
$ singularity sign image.sif 
Signing image with PGP key material
Signature created and applied to image.sif
```

Environment variable set:

```
$ export SIGN_KEY=private.pem
$ singularity sign image.sif 
Signing image with key material from 'private.pem'
Signature created and applied to image.sif
```

> 2. If the top level `test` directory is only going to contain test data, and not code for tests, would it be reasonable to use `testdata` instead? As I understand it, the go toolchain would then ignore any `.go` files in there.... which would make it a good place to 'hide' deprecated plugin examples that we may still need to test compile for a time, but don't want to keepin in `examples`, have appearing in go doc etc.
> 
> > Directory and file names that begin with "." or "_" are ignored by the go tool, as are directories named "testdata".

I wonder if it might be worth following the guidelines [here](https://github.com/golang-standards/project-layout#test), as I think the current structure was based on those recommendations to some extent? That wouldn't preclude us from having a `test/testdata` directory for the reasons you've articulated, while also supporting test _code_ (`e2e`->`test/e2e`, for example).

tri-adam:
Also going to raise a situation that just occurred to me... in the event that an image is signed _both_ with PGP and non-PGP key material:

```sh
$ singularity sign image.sif
$ singularity sign --key private.pem image.sif
```

This can't be successfully verified, since using `--key` disables PGP key material as a source:

```sh
$ singularity verify image.sif 
Verifying image: image.sif
FATAL:   Failed to verify container: integrity: key material not provided for DSSE envelope signature
```

```sh
$ singularity verify --key public.pem image.sif 
Verifying image: image.sif
FATAL:   Failed to verify container: integrity: key material not provided for PGP clear-sign signature
```

Perhaps `verify` should instead consider `--key` to be _additive_ to PGP key material sources?

dtrudg (approved):

tri-adam:
> These latter 2 points seem like they can be captured as separate issues, and addressed. So putting a ✅ on here now.

Thanks, I'll merge this. I've opened #1148 and #1150 to open up discussion on the relevant points.

PR #1135, opened 2022-11-21
oci: Add --fakeroot support to --oci mode 

### LINKED ISSUE #1035: Support '--fakeroot' with OCI launcher.
Support the `--fakeroot` CLI option via the OCI launcher flow. This will launch a container where:

* The host user is non-root.
* The apparent user id in the container is root.
* The user's configured subuid and subgid mappings are in effect in the container.

dtrudg:

Implements initial --fakeroot support for --oci mode. Mirrors behavior with --compat / --contain.

The process of accomplishing this is fairly complex. In theory, we should be able to call `crun` directly from host namespaces. We can let `crun` create a default userns and user mapping, or be able to request arbitrary explicit uid mappings (permitted by /etc/subuid /etc/subgid) to be applied.

Unfortunately this can hit bugs as this path is not well tested in `crun`. See e.g.

* https://github.com/containers/crun/issues/1077
* https://github.com/containers/crun/issues/1072

To avoid requiring the very latest `crun` with fixes, instead move to creating a userns ourselves, with a 'fakeroot' ID mapping in place, and call `crun` from that. This is the kind of flow implemented in other runtimes that call `crun`, such as `podman`.... so `crun` is well tested with it.

We can use the singularity `starter` with the `fakeroot` engine ,and a minimal config, to create the userns and id mapping. We already use it this way to perform a simple `rm` cleanup in `--fakeroot` execution of a native singularity container. This approach avoids having to implement further C/Go executables, or making large imports, such as the `reexec` + `unshare` packages from `github.com/containers/storage`

We are now, by default, calling `crun` or `runc` with a `fakeroot` setup... with host uid/gid mapped to 0 inside the userns. This is default for most OCI runtimes, however singularity default is to preserve the host id.

We need to have (fake) root inside the userns for `crun` / `runc` to work properly... so to return to the host uid in the container we insert an inner user namespace / ID mapping request into the bundle `config.json`. This reverses the mapping, i.e.

- User ID on host (1001)
- Root in outer user namespace (0)
- User ID in container (1001)

I've added a README.md in this PR, as things are rather complex:

https://github.com/sylabs/singularity/blob/9680663bb1eccbfade2de9f639e1e4be49ffb3bf/internal/pkg/runtime/launcher/oci/README.md

Note that there are likely still some rough edges, and the oci launcher package is getting toward the point where it should be split, but I don't want to add too much more to this single review.

 - Fixes #1035

### DIFF
--- a/cmd/internal/cli/oci_linux.go
+++ b/cmd/internal/cli/oci_linux.go
@@ -6,6 +6,10 @@
 package cli
 
 import (
+	"errors"
+	"os"
+	"os/exec"
+
 	"github.com/spf13/cobra"
 	"github.com/sylabs/singularity/docs"
 	"github.com/sylabs/singularity/internal/app/singularity"
@@ -150,6 +154,10 @@ var OciRunCmd = &cobra.Command{
 	PreRun:                CheckRoot,
 	Run: func(cmd *cobra.Command, args []string) {
 		if err := singularity.OciRun(cmd.Context(), args[0], &ociArgs); err != nil {
+			var exitErr *exec.ExitError
+			if errors.As(err, &exitErr) {
+				os.Exit(exitErr.ExitCode())
+			}
 			sylog.Fatalf("%s", err)
 		}
 	},
--- a/e2e/actions/oci.go
+++ b/e2e/actions/oci.go
@@ -72,7 +72,7 @@ func (c actionTests) actionOciRun(t *testing.T) {
 		},
 	}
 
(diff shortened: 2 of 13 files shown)


tri-adam (approved):
Phew, there's a lot here! Thank you very much for the added README... very helpful.

Haven't had a chance to test extensively myself, so approving based on code review. Will definitely be digging deeper early next week, and don't think there's any need to hold up merging if you're comfortable doing so.

dtrudg:
Will merge this now, as there are a couple of small PRs to knock out next, and I can branch those of main then, instead of stacking the,.

PR #1155, opened 2022-11-30
Environment Variables for Signing Key Material

### LINKED ISSUE #1148: Environment Variable for Signing Key Material
`singularity sign` now has a `--key` option to specify non-PGP key material. As suggested in https://github.com/sylabs/singularity/pull/1120#pullrequestreview-1197090012, it would be helpful to have an associated env var, and perhaps some visual feedback to make it clear to the user which source of key material is being used (https://github.com/sylabs/singularity/pull/1120#issuecomment-1330740572).

For example, without the environment variable set, using PGP key material:

```sh
$ singularity sign image.sif 
Signing image with PGP key material
Signature created and applied to image.sif
```

And with non-PGP key material provided via environment variable:

```
$ SIGN_KEY=private.pem singularity sign image.sif 
Signing image with key material from 'private.pem'
Signature created and applied to image.sif
```

And finally, with non-PGP key material provided via flag:

```
$ singularity sign --key private.pem image.sif
Signing image with key material from 'private.pem'
Signature created and applied to image.sif
```

tri-adam:

Add visual indication of key material being used by `singularity sign`. Improve `singularity sign` documentation. Define `SINGULARITY_SIGN_KEY` environment variable for '--key' flag of `singularity sign`. Simplify `singularity sign` end-to-end tests, and add coverage for `SINGULARITY_SIGN_KEY` environment variable.

 - Fixes #1148 

### DIFF
--- a/cmd/internal/cli/sign.go
+++ b/cmd/internal/cli/sign.go
@@ -7,7 +7,6 @@ package cli
 
 import (
 	"crypto"
-	"fmt"
 
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
@@ -132,7 +131,7 @@ func doSignCmd(cmd *cobra.Command, cpath string) {
 	// Set key material.
 	switch {
 	case cmd.Flag(signPrivateKeyFlag.Name).Changed:
-		fmt.Printf("Signing image with key material from '%s'\n", priKeyPath)
+		sylog.Infof("Signing image with key material from '%v'", priKeyPath)
 
 		s, err := signature.LoadSignerFromPEMFile(priKeyPath, crypto.SHA256, cryptoutils.GetPasswordFromStdIn)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/internal/cli/sign.go
```diff
@@ -131,13 +132,17 @@ func doSignCmd(cmd *cobra.Command, cpath string) {
 	// Set key material.
 	switch {
 	case cmd.Flag(signPrivateKeyFlag.Name).Changed:
+		fmt.Printf("Signing image with key material from '%s'\n", priKeyPath)
```
dtrudg, line 135: Part of me wonders if we should switch these to sylog.Infof at this point, so they can be `--quiet` or `--silence` 'ed

Arguably they aren't explicit output, but informative messages appropriate for log output rather than direct STDOUT.

dtrudg: Though perhaps deferring it til 4.0 is preferable, in case anyone is looking for these on STDOUT.

dtrudg (approved):
LGTM

cmd/internal/cli/sign.go
tri-adam: Given that we've changed the content of the console output anyways, it's difficult to imagine a scenario where switching to `INFO` messages would cause additional breakage?

Personally, I'm not sure we can/should provide any guarantee on stability of console output. In cases like this we're adding valuable feedback in the context of a new feature, and that's a good thing! Doesn't mean that one can't choose to parse it, but hopefully it's clear that approach will entail maintenance as Singularity evolves.

dtrudg (approved):

Issue #1152, opened 2022-11-29
Revocation Checks for X.509 Certificates


tri-adam:
As discussed in https://github.com/sylabs/singularity/issues/1095, the `singularity verify` command should support optional revocation checks via the Online Certificate Status Protocol (OCSP) and/or Certificate Revocation List (CRL).

@fnikolai, I know you've already taken a look at OCSP... if you want to jump in here, we'd be very grateful for the contribution.

fnikolai:
@tri-adam Sure. I 'll try to integrate it in the next couple of days. 

However, "end-to-end testing" can be an issue, in the sense that we need a properly signed certificate, with root and intermediate certificates. 

Since I do not have such a certificate, I 've tested the OCSP code by verifying Google's certificate chain.

To download the certificate chain:
```
openssl s_client -connect google.com:443 -showcerts
```

tri-adam:
Awesome. Sounds good on the approach of using Google's certificate chain... that appears to be the approach taken in the `golang.org/x/crypto/ocsp` unit tests ([ref](https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.3.0:ocsp/ocsp_test.go)).

In terms of end-to-end testing, yeah, it'd be pretty tough. We could probably hack something by implementing a local OCSP responder, and then generating a certificate that references it on the fly using code similar to what's in [gen_certs.go](https://github.com/sylabs/singularity/pull/1139/files#diff-15c61b1546f46db34a8b5b19ce7ca26aea8a2ed4b37528642c06221837f429ea). Alternatively, we could test the unhappy path... generate a certificate that references an OCSP responder that _won't_ say it's valid, and just make sure that the command errors out as expected.

fnikolai:
@tri-adam
Good news ... and bad news!

The PR is ready.
https://github.com/sylabs/singularity/pull/1160

However, I had used your verify-certificates branch as a reference, which now seems to be deleted ... and that carried some other commits :S (I admit my poor git skills).

In any case, the only files that were changed are:

    cmd/internal/cli/{verify,x509}.go
    internal/app/singularity/verify.go {required the x509.ExtKeyUsageClientAuth}

tri-adam:
Hmm, the branch hasn't been deleted... it's available here: https://github.com/tri-adam/singularity/tree/verify-certificate. You should be able to `rebase` onto it I think?

fnikolai:
[...]

PR #1160, opened 2022-12-02
Add support for OCSP

fnikolai:
This PR implements the OCSP support discussed in https://github.com/sylabs/singularity/issues/1095 and https://github.com/sylabs/singularity/issues/1152.

The reference implementation is based on RFC https://www.rfc-editor.org/rfc/rfc6960

Tests has been based on this doc.
https://akshayranganath.github.io/OCSP-Validation-With-Openssl/

Step 1: Get the server certificate
```
openssl s_client -connect www.akamai.com:443 < /dev/null 2>&1 |  sed -n '/-----BEGIN/,/-----END/p' > certificate.pem
```

Step 2: Get the intermediate certificate
```
openssl s_client -showcerts -connect www.akamai.com:443 < /dev/null 2>&1 |  sed -n '/-----BEGIN/,/-----END/p' > chain.pem
```

Step 3: Run Singularity in OCSP mode 
```
singularity verify --certificate ./certificate.pem --certificate-intermediates ./chain.pem --ocsp-verify image.sif
```

Expected outputs:

- **If communication error:** FATAL:   OCSP server is unavailable: certificate 'CN=www.akamai.com,O=Akamai Technologies\, Inc.,L=Cambridge,ST=Massachusetts,C=US': failed to contact OCSP (additional information is printed to identify which certificate has failed)

- **If OCSP OK:** FATAL:   Failed to verify container: integrity: signature object 5 not valid: dsse: verify envelope failed: Accepted signatures do not match threshold, Found: 0, Expected 1 (reasonable since the certificate is just for testing).

- **If use invalid certificates**: FATAL:   Failed to verify container: x509: certificate signed by unknown authority


cmd/internal/cli/x509.go
```diff
@@ -0,0 +1,288 @@
+// Copyright (c) 2022, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package cli
```
dtrudg, line 6: If possible, this implementation should not be in the `cmd/internal/cli` package but somewhere appropriate under `internal/pkg`.

We're trying to move implementation logic out of the `cli` package, so that it really just has flag / option handling etc.

fnikolai: I see. It shouldn't be a problem.

cmd/internal/cli/verify.go
```diff
@@ -167,6 +213,72 @@ func doVerifyCmd(cmd *cobra.Command, cpath string) {
...
+	case cmd.Flag(verifyCertificateFlag.Name).Changed:
+		cert, err := loadCertificate(certificatePath)
+		if err != nil {
+			sylog.Fatalf("Failed to load certificate: %v", err)
+		}
+		opts = append(opts, singularity.OptVerifyWithCertificate(cert))
+
+		// If the certificate is self-signed certificate, we can use the certificate itself as one of the roots.
+		// Otherwise, we need the certificate of the authorities who signed the certificate.
+		var certificateIsSelfSigned bool
```
tri-adam, line 225: I think it would be wise to tackle any self-signed certificate functionality separate from the online verification work. We probably need to discuss what _should_ happen when a self-signed certificate is specified. My gut instinct is that Singularity should either:
- Not accept a self-signed cert unless it is present in the system pool, or
- Accept it only if the user has specified this as the desired behaviour.

As an example of how this is handled elsewhere, `wget --no-check-certificate` allows a self-signed cert to be used. That being said, its documentation actually recommends the first approach for self-signed certs:
  > For self-signed/internal certificates, you should download the
           certificate and verify against that instead of forcing this insecure mode.

cmd/internal/cli/x509.go
```diff
@@ -0,0 +1,288 @@
...
+
+		pool[string(cert.SubjectKeyId)] = cert
+	}
+
+	return pool, nil
+}
+
+// certPool is needed because *x509.x509CertPool does not support listing of Certificates.
+// Without listing, it is impossible to perform Revocation Checking.
+type certPool map[string]*x509.Certificate
```
tri-adam, line 68: I'm not sure I understand why we can't do revocation listing without defining this type?

The logic in `internal/app/singularity` is calling [Certificate.Verify](https://pkg.go.dev/crypto/x509#Certificate.Verify) ([here](https://github.com/fnikolai/singularity-x509/blob/f3d5b75b0ac3023c445b246f4ceb4f3a7909cb33/internal/app/singularity/verify.go#L168)), which will return one or more certificate chains based on the certificate and pool(s) it is passed.

When `--ocsp-verify` is specified, I think what we'll need to do is add on to that by walk the certificate chain(s) and do OCSP verification (call `queryOCSP`) on each certificate in the chain. Does that make sense?

PR #1208, opened 2022-12-22
feat: oci: honor USER in image config

### LINKED ISSUE #77: Support for Dockerfile USER (--oci mode)
Some Docker containers were built with a Dockerfile containing a `USER` directive. The application was installed under the assumption that at runtime the specified user is active in the container. This is not the case with Singularity, so these containers may fail with permissions or application configuration errors.

~~SingularityCE has a ‘fakeroot engine’ that is able to configure a container run so that subuid/subgid configuration is used. This type of functionality opens the possibility of carrying through USER specifications from Docker containers, so that their payload can run as the expected username.~~

Support for Dockerfile USER should be enabled by the ability to execute images through runc as a low-level engine, through 3.11 and 4.0.

dtrudg:

When a USER is specified in the image config, and we are running in the `--oci` mode:

* If running unprivileged, ensure the inner uid / gid mapping results in the container process running as the USER, by default.
* If running privileged, run as the USER, by default.

This has been accomplished by:

* Refactoring handling of the OCI runtime config Process component, out of the bundle preparation, into the launcher code.
* Following the execution path:
  * Launcher creates a runtime spec, with configuration that is independent of the image config.
  * Bundle is created, with this minimum initial spec. Bundle creation retrieves and extracts the image, so the image config is now available.
  * Launcher gets image config from the bundle, uses it to finalize runtime spec configuration that is dependent on the image config.
  * Bundle is updated with the finalized runtime spec config.
  * Launcher executes container using the finalized bundle.

**Notes**

* Might be easier to review commit-by-commit....
  * 1st commit does refactoring of runtime config Process handling out of bundle.
  * 2nd commit adds utility functions to pull User info out of a rootfs passwd/group file in a bundle.
  * 3rd commit implements the `USER handling`.
  
**Follow up**
  
In `--oci` mode, you will now always enter a container as the `USER`, if there is one defined. The exception to this being if `--fakeroot` is specified (will be root in the container).

In practice, this is not a big issue for the unprivileged case, since `USER` in the container maps through userns / subid mappings to the host user still.... and we are trying to mirror the default behavior of OCI runtimes here.

However, it would be good to have the option to enter as a different container uid/gid.... to override `USER`, other than with `--fakeroot`.

 - Fixes #77


tri-adam (approved):
Makes sense to me.

Issue #1150, opened 2022-11-29
Verification of Multiple Signatures


tri-adam:
When `singularity verify` is called on an image with a single signature, the expected behaviour is straightforward. If the signature can be verified, all is well, and otherwise there is a problem.

When `singularity verify` is called on an image with multiple signatures, the expected behaviour is less clear. It could be:

1. Verification is successful if all signatures are verified. This is the current behaviour.
1. Verification is successful if one or more (but possibly not all) signatures were verified.

Although the first option is the current behaviour, I would argue that it likely isn't the user's expected behaviour in all cases. My intent may be "verify this image has been signed by a particular authority". In this case, I could ensure I have the public key material for the relevant authority, and call `verify`. Unfortunately, if the image contains other signatures that are not relevant to me, and I don't have access the the required public key material, verification fails.

To make matters worse, in the above use case, `verify` doesn't have a way for me to specify the relevant authority. The suggestion in https://github.com/apptainer/apptainer/issues/462 goes part way to helping with this by displaying "verified" or "not verified" for each signature, but still requires the user to (visually or programmatically) parse the output to determine whether a signature from the relevant authority can be verified.

In addition, `singularity verify` now has a `--key` option to specify non-PGP key material, which introduces yet another complexity. The user expectation of `singularity verify --key` may very well be "verify this image can be verified with this key material", but at present the semantics are actually "verify all signatures in this image can be verified with this key material". This means that verifying an image signed with both PGP and non-PGP key material actually isn't possible, since there is no way to specify multiple key material sources:

```sh
$ singularity sign image.sif
$ singularity sign --key private.pem image.sif
```

This can't be successfully verified, since using `--key` disables PGP key material as a source:

```sh
$ singularity verify image.sif 
Verifying image: image.sif
FATAL:   Failed to verify container: integrity: key material not provided for DSSE envelope signature
```

```sh
$ singularity verify --key public.pem image.sif 
Verifying image: image.sif
FATAL:   Failed to verify container: integrity: key material not provided for PGP clear-sign signature
```

As a proposal, I suggest the following semantics:

1. `singularity verify image.sif` maintains the current semantics, i.e. all signatures in `image.sif` are verified using PGP public key material.
2. `singularity verify --key public.pem image.sif` verifies `image.sif` can be verified by `public.pem`. Same if key material is provided by environment variable (ex. `VERIFY_KEY=public.pem singularity verify image.sif `).
[...]

PR #1213, opened 2022-12-26
Add support for OCSP verification

fnikolai:
Signed-off-by: Fotis Nikolaidis <nikolaidis.fotis@gmail.com>

Add support for Online Revocation Checks using the OCSP protocols. 

 - Fixes #https://github.com/sylabs/singularity/issues/1152

#### Testing & Limitations 
Since there is no officially signed certificate chain for Singularity, the validation of OCSP is done:
1) Using the self-signed certificates. However, e2e testing fails because the root certificate is signed using Ed25519 (see the issue discussion)

2) Using third-party certificate chain (AKAMAI). OCSP successfully validates the certificate, but the signature verification fails (since they are not useful for signing the image).


e2e/verifyocsp/responder/standalone/add_cert_to_index.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
dtrudg, line 1: If you can, please use `#!/bin/sh` and do a shellcheck for POSIX sh compatibility - we are packaged for alpine where `bash` cannot be counted upon.

I can tidy this up post merge if that's a pain.

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@
 
 ### New features / functionalities
 
+- Support for online verification checks of x509 certificates using OCSP protocol.
```
dtrudg, line 36: Please add to the bottom of the list - for convenience reasons we try to keep things as chronological as possible here. If we re-order things in the release notes, that's handled separately.

cmd/internal/cli/verify.go
```diff
@@ -210,6 +222,13 @@ func doVerifyCmd(cmd *cobra.Command, cpath string) {
 		if err != nil {
 			sylog.Fatalf("Failed to load certificate: %v", err)
 		}
+
+		// reject self-signed certificates.
```
dtrudg, line 226: I'm not sure about this being globally applied here. There are very valid reasons to use a self-signed certificate (outside of OCSP)... and if we block their use here I'd expect we will get issues raised by people testing things.

e2e/suite.go
```diff
@@ -92,6 +93,7 @@ var e2eGroups = map[string]testhelper.Group{
 	"SECURITY":   security.E2ETests,
 	"SIGN":       sign.E2ETests,
 	"VERIFY":     verify.E2ETests,
+	"OCSPVERIFY": verifyocsp.E2ETests,
```
dtrudg, line 96: Because the new functionality is entirely part of the `verify` command, I'd prefer these tests are added to the `VERIFY` group / package, rather than a new group created.

We usually want to keep things for a command together, except when there are specific reasons to separate them, or they test functionality that is implemented across multiple commands.

internal/app/singularity/verify_ocsp.go
```diff
@@ -0,0 +1,237 @@
+// Copyright (c) 2022, Sylabs Inc. All rights reserved.
+// Copyright (c) 2020-202, ICS-FORTH.  All rights reserved.
```
dtrudg, line 2:
```suggestion
// Copyright (c) 2020-2022, ICS-FORTH.  All rights reserved.
```

e2e/verifyocsp/verify-ocsp.go
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) 2019-2022, Sylabs Inc. All rights reserved.
```
dtrudg, line 1: This should be the appropriate copyright statement for you, and dated this year.

e2e/verifyocsp/responder/ocsp_responder.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2022, Sylabs Inc. All rights reserved.
```
dtrudg, line 1: This should be the appropriate copyright statement for you.

internal/app/singularity/verify.go
```diff
@@ -167,10 +179,28 @@ func (v verifier) getOpts(ctx context.Context, f *sif.FileImage) ([]integrity.Ve
...
+		// verify that the leaf certificate is not tampered and that is adequate for signing purposes.
+		chain, err := verifyCertificate(c, v.intermediates, v.roots)
+		if err != nil {
 			return nil, err
 		}
 
+		// Verify that the certificate is issued by a trustworthy CA (i.e the certificate chain is not revoked or expired).
+		if v.ocsp {
+			if len(chain) != 1 {
+				return nil, errors.New("Unhandled OCSP condition")
```
dtrudg, line 191:
```suggestion
				return nil, fmt.Errorf("unhandled OCSP condition, chain length %d != 1", len(chain))
```

internal/app/singularity/verify.go
```diff
@@ -167,10 +179,28 @@ func (v verifier) getOpts(ctx context.Context, f *sif.FileImage) ([]integrity.Ve
...
+			}
+
+			ocspErr := OCSPVerify(chain[0]...)
+			if ocspErr != nil {
+				// TODO: We need to decide whether this should be strict or permissive.
+				return nil, ocspErr
+			}
+
+			sylog.Infof("OCSP validation has passed")
```
dtrudg, line 200: Should this be part of the verification message output that goes to STDOUT, rather than a log message to STDERR?

Another way of thinking about it... if someone uses `--quiet`... should this still appear?

internal/app/singularity/verify_ocsp.go
```diff
@@ -0,0 +1,237 @@
...
+
+/*
+Online Certificate Status Protocol - OCSP
+
+OCSP responder is used to provide real-time verification of the revocation status of an X.509 certificate.
+RFC: https://www.rfc-editor.org/rfc/rfc6960
+*/
+
+const (
+	PKIXOCSPNoCheck = "1.3.6.1.5.5.7.48.1.5"
```
dtrudg, line 31: Would be nice to have a bit more of a note what this is... for anyone who hasn't grepped the OCSP docs.

internal/app/singularity/verify_ocsp.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func validateCertificate(cert *x509.Certificate, pool map[string]*x509.Certificate) error {
+	if len(cert.AuthorityKeyId) == 0 || string(cert.SubjectKeyId) == string(cert.AuthorityKeyId) {
+		sylog.Infof("skip self-signed certificate (%s)", cert.Subject.String())
+
+		return nil
+	}
+
+	/*---------------------------------------------------
```
dtrudg, line 63: Please use single-line `//` comments here, and below, to match style. No need for header bars either. I don't think we have any multi-line comments in our Go code.

internal/app/singularity/verify_ocsp.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func downloadCertsFromURLs(urls ...string) (map[string]*x509.Certificate, error) {
+	certs := map[string]*x509.Certificate{}
+
+	for _, certURL := range urls {
+		sylog.Infof("Downloading certificate from %s", certURL)
+
+		//nolint:gosec
+		// Alternative:  GOSEC=gosec -quiet -exclude=G104,G107
```
dtrudg, line 211: If you disable a gosec rule, please explain why it's necessary / safe.

dtrudg (dismissed):
Hi @fnikolai, and many thanks again for your continued work on this!

I've left some more general comments on the code. I'm sure that @tri-adam will be along soon to go over the implementation in more detail.

Let me know if there are any questions.

e2e/verifyocsp/thirdparty-certificates/download_chain.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Step 1: Get the leaf certificate
+openssl s_client -connect www.akamai.com:443 < /dev/null 2>&1 |  sed -n '/-----BEGIN/,/-----END/p' > leaf.pem
+
+# Step 2: Get the intermediate certificate
+openssl s_client -showcerts -connect www.akamai.com:443 < /dev/null 2>&1 |  sed -n '/-----BEGIN/,/-----END/p' > intermediate.pem
```
tri-adam, line 7: I don't think the `www.akamai.com` cert _should_ pass validation, since it's not valid for code signing, only TLS:

```
$ openssl x509 -in e2e/verifyocsp/thirdparty-certificates/leaf.pem -text -noout
...
            X509v3 Extended Key Usage: 
                TLS Web Server Authentication, TLS Web Client Authentication
```

I'm not sure off the top of my head where we could easily obtain a valid code signing chain. By definition, reaching out to a TLS server will get us a certificate for that rather than code signing.

To move forward, I propose we remove the third party certificate stuff for now, and the e2e test case that uses it, and address that separately. Does that sound like a plan?

internal/app/singularity/verify_ocsp.go
```diff
@@ -0,0 +1,237 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "download cert error")
+		}
+
+		// if that does not work either, just abort
+		issuer, exists = missingCerts[string(cert.AuthorityKeyId)]
+		if !exists {
+			return errors.Errorf("cannot find issuer '%s'", cert.Issuer)
+		}
+	}
```
tri-adam, line 81: By definition, certificate N in the chain was issued by certificate N+1 all the way to the root, so I don't think we should ever need to download a certificate during OCSP verification?

internal/app/singularity/verify_ocsp.go
```diff
@@ -0,0 +1,237 @@
...
+		// RFC-6960 Section: 4.2.2.2.1
+		for _, extension := range cert.Extensions {
+			if extension.Id.String() == PKIXOCSPNoCheck {
+				goto skipOCSPVerification
+			}
+		}
+
+		// avoid loops cause by misconfigured OCSP responders
+		if string(cert.SubjectKeyId) == string(ocspCertificate.SubjectKeyId) {
+			return nil
```
tri-adam, line 105: I wonder if an error should be returned here?

internal/app/singularity/verify_ocsp.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+
+		// make sure that the OCSP server is trustworthy
+		if err := validateCertificate(ocspCertificate, pool); err != nil {
+			return errors.Wrapf(err, "cannot verify OCSP server's certificate")
+		}
```
tri-adam, line 111: I believe this only validates that the cert isn't revoked? I believe we need to call [Certificate.Verify](https://pkg.go.dev/crypto/x509#Certificate.Verify) as well to ensure the certificate is valid?

internal/app/singularity/verify_ocsp.go
```diff
@@ -0,0 +1,237 @@
...
+
+	skipOCSPVerification:
+	}
+
+	return nil
+}
+
+func queryOCSP(cert, issuer *x509.Certificate) (needsValidation *x509.Certificate, err error) {
+	logrus.Warnf("cert:[%s] issuer:[%s]", cert.Subject.String(), issuer.Subject.String())
```
tri-adam, line 120: Should use `sylog` here:

```suggestion
	sylog.Debugf("cert:[%s] issuer:[%s]", cert.Subject.String(), issuer.Subject.String())
```

internal/app/singularity/verify_ocsp.go
```diff
@@ -0,0 +1,237 @@
...
+	 * Retrieve the CA who issued the certificate in question.
+	 *---------------------------------------------------*/
+
+	// firstly, look for the issuer in the pool of certificates.
+	issuer, exists := pool[string(cert.AuthorityKeyId)]
+	if !exists {
+		// if not found anywhere locally, try to download it
+		missingCerts, err := downloadCertsFromURLs(cert.IssuingCertificateURL...)
+		if err != nil {
+			return errors.Wrapf(err, "download cert error")
```
tri-adam, line 73: Applicable in several places... use wrapped errors from the standard library, ex:

```suggestion
			return fmt.Errorf("download cert error: %w", err)
```

tri-adam (commented):
@fnikolai, looking good! Left some comments/questions. Thank you!

e2e/verifyocsp/thirdparty-certificates/download_chain.sh
fnikolai: This part follows the discussion on
https://github.com/sylabs/singularity/issues/1152#issuecomment-1332183114

For passing the validation I had to append the `x509.ExtKeyUsageClientAuth`
https://github.com/sylabs/singularity/blob/3081ce65c3dae52d6238e310bfe0dc73ef67725c/internal/app/singularity/verify.go#L163-L169

I could remove it, but then there is no way for testing that OCSP actually works. 

The existing certificate chain fails early (due to the ed25519 issue), which makes it impossible to test the OCSP codepath to the end. Alternatively, we could create an RSA-only certificate chain and let OCSP complete the validation. 

How would you like to proceed?

internal/app/singularity/verify_ocsp.go
fnikolai: This part is used to validate the OCSP's certificate -- not for the code signing certificate. 

At least, this is my interpretation of `4.2.2.2  Authorized Responders` on https://www.ietf.org/rfc/rfc2560.txt

internal/app/singularity/verify_ocsp.go
fnikolai: Restructured the code with more appropriate handling of such cases.

internal/app/singularity/verify_ocsp.go
fnikolai: We could add it, but I don't think that is mentioned in the [RFC](https://www.ietf.org/rfc/rfc2560.txt). 

Plus, it will be recursively validated by the OCSP server.

dtrudg:
Thanks for the updates @fnikolai - it'll likely get re-reviewed toward the end of this week / early next week as people return from holidays.

internal/app/singularity/verify_ocsp.go
tri-adam: > We could add it, but I don't think that is mentioned in the [RFC](https://www.ietf.org/rfc/rfc2560.txt).
> 
> Plus, it will be recursively validated by the OCSP server.

I believe `§2.2 Response` does mandate this?

> All definitive response messages SHALL be digitally signed. The key used to sign the response MUST belong to one of the following:
> 
>    -- the CA who issued the certificate in question
>    -- a Trusted Responder whose public key is trusted by the requester
>    -- a CA Designated Responder (Authorized Responder) who holds a specially marked certificate issued directly by the CA, indicating that the responder may issue OCSP responses for that CA

In the last case, [ocsp.ParseResponseForCert](https://pkg.go.dev/golang.org/x/crypto/ocsp#ParseResponseForCert) calls [Response.CheckSignatureFrom](https://pkg.go.dev/golang.org/x/crypto/ocsp#Response.CheckSignatureFrom), which validates that the supplied issuer signed the certificate, but does not validate that `resp.Certificate` is authorized to issue OCSP responses as indicated in §4.2.2.2. In other words, if `resp.Certificate` isn't `nil`, I believe we need to validate that it includes an an `ExtendedKeyUsage` of `id-ad-ocspSigning`?

Not sure if that's clear... this stuff is complicated!

internal/app/singularity/verify_ocsp.go
tri-adam: > This part is used to validate the OCSP's certificate -- not for the code signing certificate.

I think it's used for both? It's called from `OCSPVerify` [here](https://github.com/fnikolai/singularity-x509/blob/49bb33b0f81c84517d679e18e8324581fa93dd7d/internal/app/singularity/verify_ocsp.go#L48-L54), which is validating the code signing certificate?

> At least, this is my interpretation of `4.2.2.2 Authorized Responders` on https://www.ietf.org/rfc/rfc2560.txt

My understanding from that section is there are three cases to consider:

1. _Matches a local configuration of OCSP signing authority for the certificate in question:_ In this case, the responder certificate would be contained in the local configuration, and 
2. _Is the certificate of the CA that issued the certificate in question:_ In this case, the responder certificate is the same as the CA of the certificate in question, and included in the chain.
4. _Includes a value of `id-ad-ocspSigning` in an ExtendedKeyUsage extension and is issued by the CA that issued the certificate in question:_ In this case, the responder certificate must be signed by the same CA as the certificate in question, and thus can be validated using the same intermediate and root pools.

In all three cases, it's unnecessary to download an issuer (CA) certificate? It's either in local configuration (we're not implementing this currently), or present as a CA in the chain. Am I missing a case where the issuer of a OCSP certificate needs to be downloaded?

fnikolai: I was thinking about the case of an OCSP responder being signed by intermediate `ocspSigning` certificates that have the same Root CA as the `codeSigning` certificates. In that case, we had to download the chain of intermediate `ocspSigning` certificates.

In other words, my assumption was the OCSP Responder can be indirectly signed by the RootCA. But this case is excluded by `2.2  Response`:
```
The key used to sign the response MUST belong to one of the following:
   -- the CA who issued the certificate in question
   -- a Trusted Responder whose public key is trusted by the requester
   -- a CA Designated Responder (Authorized Responder) who holds a specially marked certificate issued directly by the CA, indicating that the responder may issue OCSP responses for that CA.
```

So, I 'm removing the corresponding code.

internal/app/singularity/verify_ocsp.go
fnikolai: Aham. I see two parallel threads here.
- How do we know that the OCSP response has not been tampered with?
- How do we know that the OCSP responder is trustworthy? 

== Response Integrity ==
As you said, [ocsp.ParseResponseForCert](https://pkg.go.dev/golang.org/x/crypto/ocsp#ParseResponseForCert) calls [Response.CheckSignatureFrom](https://pkg.go.dev/golang.org/x/crypto/ocsp#Response.CheckSignatureFrom), which validates that the supplied issuer signed the certificate. 

If the returned certificate is not `nil`, the `verify_ocsp.go` goes through the logic for `Authorized Responders`

https://github.com/sylabs/singularity/blob/49bb33b0f81c84517d679e18e8324581fa93dd7d/internal/app/singularity/verify_ocsp.go#L91-L92

== Authorized Responders ==
The logic of `§4.2.2.2.` is already implemented and documented in the code.

In particular for the  `ExtendedKeyUsage`: 
https://github.com/sylabs/singularity/blob/49bb33b0f81c84517d679e18e8324581fa93dd7d/internal/app/singularity/verify_ocsp.go#L112-L115


Is that ok, or do you mean something different?

tri-adam: Cool, yes, I think we're good on both those counts now!

internal/app/singularity/verify.go
```diff
@@ -148,17 +160,17 @@ func newVerifier(opts []VerifyOpt) (verifier, error) {
...
 // more chains from c to a certificate in roots, using certificates in intermediates if needed.
 // This function does not do any revocation checking.
-func verifyCertificate(c *x509.Certificate, intermediates, roots *x509.CertPool) error {
+func verifyCertificate(c *x509.Certificate, intermediates, roots *x509.CertPool) (chains [][]*x509.Certificate, err error) {
 	opts := x509.VerifyOptions{
 		Intermediates: intermediates,
 		Roots:         roots,
 		KeyUsages: []x509.ExtKeyUsage{
 			x509.ExtKeyUsageCodeSigning,
+			x509.ExtKeyUsageClientAuth,
```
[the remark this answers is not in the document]
tri-adam: This is the one remaining issue that I can see. We should only be accepting certificates that are valid for code signing, not TLS authentication.

tri-adam (dismissed):
This is the last remaining issue that I can see.

I'm conscious that the E2E tests are using a cert chain that is from a TLS client/server, and that removing this will break that test. For now, I think it might be best to remove that test, or assert that it fails, so this work can be merged.

fnikolai:
> ... For now, I think it might be best to remove that test, or assert that it fails, so this work can be merged.

@tri-adam `x509.ExtKeyUsageClientAuth` is removed and the proper assertion is added to the test.

> Looks like this will be mergeable soon, @tri-adam / @fnikolai ? I'll add it to the 3.11 milestone now on that assumption.
> 
> Please let me know if I'm wrong on this. Thanks!

@dtrudg I believe it is now ready to be merged.

dtrudg:
> @dtrudg I believe it is now ready to be merged.

There's a conflict on `go.mod` due to the changes that have happened on the `main` branch recently. If you are able, could you rebase onto `main`? Otherwise if the box is checked to allow maintainers to modify the PR branch, I can take care of that after @tri-adam indicates he's happy.

Thanks!

fnikolai:
@dtrudg done

dtrudg:
@fnikolai - LGTM. I'll do a final quick sanity check for my benefit then merge, as tri-adam is travelling.

Many thanks for your efforts on this, and patience working through the reviews etc. over the holiday period.

dtrudg (approved):

fnikolai:
> > @dtrudg I believe it is now ready to be merged.
> 
> There's a conflict on `go.mod` due to the changes that have happened on the `main` branch recently. If you are able, could you rebase onto `main`? Otherwise if the box is checked to allow maintainers to modify the PR branch, I can take care of that after @tri-adam indicates he's happy.
> 
> Thanks!

It's my pleasure to jump in. Looking forward to the next contribution. 

PS, I believe https://github.com/sylabs/singularity/issues/1152 and perhaps https://github.com/sylabs/singularity/issues/1150 can be closed.

PR #1222, opened 2023-01-04
oci: Use cache more efficiently for --oci mode

### LINKED ISSUE #1221: Minimal --oci mode image caching
At present, when an OCI image is used in `--oci` mode, layer blobs are pulled and cached, before an oci-layout is created which will then be extracted. We always have 2 instances of a containers/image Copy operation.

~~OCI -> SIF conversion caches a converted image. We could do simple caching of an oci-archive for an image with a specific hash in order to remove one of the Copy operations and speed things up a bit.~~

Because the OCI cache is a transparent wrapper, we can just pull through it when the cache is enabled. No need for any caching of an archive etc.

dtrudg:

When the cache is enabled, and `--oci` mode is used, extract the image directly from the cache oci-layout, rather than copying into a temporary oci-layout first.

In addition, within the transparent caching functions, check if an image manifest is present before performing a copy into the cache. Although the copy is effectively a no-op if the image is present, this avoids some overhead.

 - Fixes #1221

### DIFF
--- a/internal/pkg/build/oci/oci.go
+++ b/internal/pkg/build/oci/oci.go
@@ -70,7 +70,12 @@ func (t *ImageReference) newImageSource(ctx context.Context, sys *types.SystemCo
 		return nil, err
 	}
 
-	// First we are fetching into the cache
+	// Check if the image is in the cache layout already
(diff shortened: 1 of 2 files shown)


tri-adam (approved):
LGTM

PR #1355, opened 2023-02-23
fix: avoid incorrect error with --fakeroot --net --network=fakeroot

### LINKED ISSUE #1352: "ERROR:   Network fakeroot is not permitted for unprivileged users." when using network with --fakeroot, but command succeeds
**Version of Singularity**
```
$ /usr/bin/singularity --version
singularity-ce version 3.11.0-1
```

**Describe the bug**

When trying to run singularity in new network namespace with fakeroot, the command succeeds, but it outputs an error:

```
ERROR:   Network fakeroot is not permitted for unprivileged users.
```

**To Reproduce**

```
$ singularity -s exec --fakeroot --net --network=fakeroot sles12sp5.sif ip -o -4 addr show scope global
ERROR:   Network fakeroot is not permitted for unprivileged users.
2: eth0    inet 10.23.0.4/16 brd 10.23.255.255 scope global eth0\       valid_lft forever preferred_lft forever
```

Without `--fakeroot` the command fails (as expected) and says that `network requires root or --fakeroot`.
```
$ singularity -s exec --net --network=fakeroot sles12sp5.sif ip -o -4 addr show scope global
ERROR:   Network fakeroot is not permitted for unprivileged users.
FATAL:   container creation failed: network requires root or --fakeroot, non-root users can only use --network=none unless permitted by the administrator
```

If in `/etc/singularity/singularity.conf` I configure `allow net networks = fakeroot`, the error is not shown.

```
$ grep 'allow net' /etc/singularity/singularity.conf
#allow net users = gmk, singularity
#allow net groups = group1, singularity
# groups listed in the allow net users / allow net groups directives. Thus feature
allow net networks = fakeroot

$ singularity -s exec --fakeroot --net --network=fakeroot sles12sp5.sif ip -o -4 addr show scope global
2: eth0    inet 10.23.0.9/16 brd 10.23.255.255 scope global eth0\       valid_lft forever preferred_lft forever
```

**Expected behavior**

- If network fakeroot should **not** be allowed for unprivileged users (even if running under `fakeroot`), then the command above should fail.
- If network fakeroot **should** be allowed for unprivileged users (when running under `fakeroot`), then we should not see this error.

**Additional context**
It could also be nice if the behavior is described in the documentation.

dtrudg:

Due to incorrect logic / ordering of checks in prepareNetworkSetup, the combination of `--fakeroot --net --network=fakeroot` generated a false error stating that the fakeroot network was not permitted.

However, execution continued, and the fakeroot network was (correctly) used.

This is a minimal PR to correct the ordering of checks, such that `--fakeroot --net --network=fakeroot` doesn't give a false error.

The code handling checks for permitted networks is messy, and still needs a proper refactor. There is overriding of the `--network` setting for `--fakeroot` in the CLI->launcher layer as well, due to way in which the default network is implemented, via flag default value.

In addition, there doesn't seem to be a good reason as to why `allowed net networks` shouldn't apply when `--fakeroot` is being used. This can be tackled in the deferred refactor.

 - Fixes #1352 

### DIFF
--- a/internal/pkg/runtime/engine/singularity/container_linux.go
+++ b/internal/pkg/runtime/engine/singularity/container_linux.go
@@ -2290,7 +2290,6 @@ func (c *container) prepareNetworkSetup(system *mount.System, pid int) (func(con
 		sessionNetNs = "/netns"
 	)
 
-	fakeroot := c.engine.EngineConfig.GetFakeroot()
 	net := c.engine.EngineConfig.GetNetwork()
 
 	// If we haven't requested a network namespace, or we have but with no config, we are done here
@@ -2298,10 +2297,21 @@ func (c *container) prepareNetworkSetup(system *mount.System, pid int) (func(con
 		return nil, nil
 	}
 
-	// Otherwise start checking what's permitted for the current user
+	// In fakeroot mode only permit the `fakeroot` CNI config, overriding any other request.
 	euid := os.Geteuid()
+	fakeroot := c.engine.EngineConfig.GetFakeroot()
+	forceFakerootNet := false
+	if fakeroot && euid != 0 && net != fakerootNet {
+		sylog.Warningf("Only --network=%s is permitted in --fakeroot mode. You requested '%s'.", fakerootNet, net)
+		sylog.Warningf("Overriding with --network=%s", fakerootNet)
+	}
+	if fakeroot && euid != 0 {
+		forceFakerootNet = true
+		net = fakerootNet
+	}
+
 	allowedNetUnpriv := false
-	if euid != 0 {
+	if euid != 0 && !forceFakerootNet {
 		// Is the user permitted in the list of unpriv users / groups permitted to use CNI?
 		allowedNetUser, err := user.UIDInList(euid, c.engine.EngineConfig.File.AllowNetUsers)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


wobito (approved):

internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -2290,18 +2290,28 @@ func (c *container) prepareNetworkSetup(system *mount.System, pid int) (func(con
...
 
-	// Otherwise start checking what's permitted for the current user
+	// In fakeroot mode only permit the `fakeroot` CNI config, overriding any other request.
 	euid := os.Geteuid()
+	fakeroot := c.engine.EngineConfig.GetFakeroot()
+	forceFakerootNet := false
+	if fakeroot && euid != 0 && net != fakerootNet {
+		sylog.Warningf("Only --network=%s is permitted in --fakeroot mode. You requested '%s'.", fakerootNet, net)
+		sylog.Warningf("Overriding with --network=%s", fakerootNet)
+	}
```
DrDaveD, line 2307: Instead of duplicating 2 of the 3 checks in this if statement with the following if statement, move this code to be indented inside the following if statement and only do the 3rd check.

internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -2314,7 +2324,11 @@ func (c *container) prepareNetworkSetup(system *mount.System, pid int) (func(con
 		// Is/are the requested network(s) in the list of networks allowed for unpriv CNI?
 		allowedNetNetwork := false
 		for _, n := range strings.Split(net, ",") {
-			allowedNetNetwork = slice.ContainsString(c.engine.EngineConfig.File.AllowNetNetworks, n)
+			// Allowed in singularity.conf
```
DrDaveD, line 2327: change singularity.conf to apptainer.conf

PR #1399, opened 2023-03-03
e2e: improve ImageVerify efficiency and replace defer-based cleanups

### LINKED ISSUE #1106: e2e: ImageVerify is inefficient
**Type of issue**

technical debt

**Description of issue**

`e2e.ImageVerify` invokes `singularity exec` 8 times, and `singularity inspect` twice to check a built image.

It should be possible to invoke each only once, performing all the necessary checks in a single container invocation.

See: https://github.com/sylabs/singularity/blob/09789989c4451eb93ea42c916e1966542634248d/e2e/internal/e2e/imageverify.go#L22

preminger:

There are two kinds of changes here. First, the number of times `singularity exec` and `singularity inspect` are invoked from the `ImageVerify()` function has been reduced from 8 and 2, respectively, to 1 and 1, respectively.

Second, and relatedly, code that invokes `ImageVerify()` and used `defer f()` statements for cleanup has been changed to use `t.Cleanup()` statements that, crucially, check for `!t.Failed()` before performing the actual cleanup operations. This is a good idea in general, since suppressing cleanups when a test fails allows for more straightforward manual inspection of the fail state. This is a change we should eventually make throughout our test suites. But in this case it's especially important, because the consolidation of `singularity exec` and `singularity inspect` invocations means that error messages, should they occur, will be less detailed than before, and manual inspection of the fail state will be especially important. (This is a trade-off between test efficiency and the quality of automated error messages, where, in this case, we have decided in favor of the former.)

 - Fixes #1106

### DIFF
--- a/e2e/internal/e2e/imageverify.go
+++ b/e2e/internal/e2e/imageverify.go
@@ -25,14 +25,14 @@ func (env TestEnv) ImageVerify(t *testing.T, imagePath string) {
 		AsSubtest("BasicIntegrityTests"),
 		WithProfile(UserProfile),
 		WithCommand("exec"),
-		WithArgs(imagePath, "/bin/sh", "-c", "( false ) || "+
-			"( test -f /.singularity.d/runscript -a "+
-			"-f /.singularity.d/env/01-base.sh -a "+
-			"-f /.singularity.d/actions/shell -a "+
-			"-f /.singularity.d/actions/exec -a "+
-			"-f /.singularity.d/actions/run -a "+
-			"-L /environment -a "+
-			"-L /singularity )"),
+		WithArgs(imagePath, "/bin/sh", "-c",
+			"test -f /.singularity.d/runscript -a "+
+				"-f /.singularity.d/env/01-base.sh -a "+
+				"-f /.singularity.d/actions/shell -a "+
+				"-f /.singularity.d/actions/exec -a "+
+				"-f /.singularity.d/actions/run -a "+
+				"-L /environment -a "+
+				"-L /singularity"),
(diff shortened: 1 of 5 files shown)


e2e/internal/e2e/imageverify.go
```diff
@@ -20,84 +20,41 @@ import (
...
 			expectOut: "1.0",
 		},
 	}
 
-	// Verify the label partition
-	for _, tt := range tests {
-		verifyOutput := func(t *testing.T, r *SingularityCmdResult) {
+	verifyOutput := func(t *testing.T, r *SingularityCmdResult) {
+		for _, tt := range tests {
 			jsonOut, err := jsonparser.GetString(r.Stdout, tt.jsonPath...)
```
dtrudg, line 58: We should be able to bring back the named subtests here, by wrapping the repeated statements with...

`t.Run(tt.name, func(t *testing.T) {`

There are only 2 now ... so it might not seem worth it, but really we should add a lot more checks on the labels at some point.

dtrudg (changes requested):
Looks pretty good to me. Pleas just either...

* Remove the commented `name` lines and field in the struct -or-
* Implement named subtests for checking the labels, like I mentioned.

I don't think it matters which you go for at this point... but we generally don't want to leave commented code unless it's for a temporary situation, explained in the comment.

Thanks!

dtrudg:
Sorry for butchering the changing of the title of the PR too....

We generally prefix stuff that is restricted to testing code with `e2e:` or `test:` so it's easy to pick out when reviewing the git history.

dtrudg (approved):

PR #1415, opened 2023-03-06
928 delete private key using cli

### LINKED ISSUE #928: Delete private key using CLI
It is currently possible to remove the public key from the local Singularity keyring. Can this feature be added to remove the private keyring as well? 

Deleting the files associated with the keyring seems like a brute-force method and does not take into consideration management of multiple keys.

 ~/.singularity/sypgp/pgp-secret 
 ~/.singularity/sypgp/pgp-public

There is a placeholder for this type of effort, to make the design and implementation more user friendly and expressive.
https://github.com/sylabs/singularity/issues/79

preminger:

Added `--secret` to the `remove` subcommand for keys. Also added `--private` as a synonym of `--secret`, as it's arguably more common to talk about 'public' / 'private', than it is to use 'secret' (despite e.g. GPG using 'secret' in commands).

For the sake of consistency, added `-p`/`--private` synonyms to `key list` and `key export`, as well.

Added e2e testing for  various conditions involving `key remove`. Also added `--private` condition to `key remove` test in `globalKeyring` (which should fail, since `key remove --private --global` is an illicit combination).

 - Fixes #928

### DIFF
--- a/cmd/internal/cli/key_export.go
+++ b/cmd/internal/cli/key_export.go
@@ -29,7 +29,17 @@ var keyExportSecretFlag = cmdline.Flag{
 	DefaultValue: false,
 	Name:         "secret",
 	ShortHand:    "s",
-	Usage:        "export a secret key",
+	Usage:        "export a secret key (synonym for --private)",
+}
+
+// -p|--private
+var keyExportPrivateFlag = cmdline.Flag{
+	ID:           "keyExportPrivateFlag",
+	Value:        &secretExport,
+	DefaultValue: false,
+	Name:         "private",
+	ShortHand:    "p",
+	Usage:        "export a private key (synonym for --secret)",
 }
 
 // -a|--armor
@@ -45,6 +55,7 @@ var keyExportArmorFlag = cmdline.Flag{
 func init() {
 	addCmdInit(func(cmdManager *cmdline.CommandManager) {
 		cmdManager.RegisterFlagForCmd(&keyExportSecretFlag, KeyExportCmd)
+		cmdManager.RegisterFlagForCmd(&keyExportPrivateFlag, KeyExportCmd)
 		cmdManager.RegisterFlagForCmd(&keyExportArmorFlag, KeyExportCmd)
 	})
 }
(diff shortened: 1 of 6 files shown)


pkg/sypgp/sypgp.go
```diff
@@ -461,6 +483,28 @@ func removeKey(list openpgp.EntityList, fingerprint string) openpgp.EntityList {
 	return nil
 }
 
+// RemovePrivKey will delete a secret key matching toDelete
+func (keyring *Handle) RemovePrivKey(toDelete string) error {
+	// read all the local private keys
```
dtrudg, line 488: We need to error out with an appropriate message if there is an attempt to remove a private key from the global keyring, as the global keyring only holds public keys (see the appendPrivateKey)

cmd/internal/cli/key_export.go
```diff
@@ -29,7 +29,17 @@ var keyExportSecretFlag = cmdline.Flag{
 	DefaultValue: false,
 	Name:         "secret",
 	ShortHand:    "s",
-	Usage:        "export a secret key",
+	Usage:        "export a secret key (synonym for --private)",
+}
+
+// -p|--private
```
dtrudg, line 35: I just had a look back at the apptainer implementation, where they added some similar functionality back about a year ago.

https://github.com/apptainer/apptainer/pull/555
https://github.com/apptainer/apptainer/pull/572

There, they added `--public / -p` `--both / -b` flags. The `--public` flag isn't needed here as it's redundant when preserving the default of removing a public key. I don't think we need `--both` at this point, either.

However.. given we will likely merge this into a 3.11 version.. let's drop the `-p` short form for `--private`, so there isn't confusion.

We can properly address this stuff for 4.0 where we can make warranted breaking changes to the CLI, and be clear about any CLI compatibility issues vs apptainer.

dtrudg (changes requested):
A few small changes requested, otherwise looks great, fitting in with the patterns that are in place in this code.

Sorry about the `--private / -p` issue... I should have recalled the apptainer decision on `-p` sooner.

dtrudg (approved):
LGTM :+1:

PR #1459, opened 2023-03-20
oci: implemented CDI device mapping

### LINKED ISSUE #1394: Add CDI --device support to --oci mode

SingularityCE doesn't currently support the new CDI standard for making hardware devices available in containers.

The `--oci` experimental runtime mode currently exposes only a naive binding approach to add devices and libraries to a container, via the `--nv` and `--oci` flags.

The `--nv` and `--oci` approach cannot support a range of functionality that is valuable, such as masking specific GPUs, introducing only subsets of device functionality into a container etc. In addition it is vendor specific, but we'd like to support e.g. Intel GPUs #1094 

The experimental `--oci` runtime mode should support a `--device` flag, allowing the user to request a device is made available in the container using a CDI configuration.

**Additional context**

Because `--oci` mode is based on OCI runtime spec generation, and invocation of `runc` / `crun`, it should be simple to implement support via spec modification carried out in the CDI module - https://github.com/container-orchestrated-devices/container-device-interface/tree/main/pkg/cdi

preminger:

Added a `--device` flag to the experimental `--oci` runtime mode, allowing the user to make a device available in the contained using a CDI configuration.

 - Fixes #1394

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/containerd/containerd v1.7.0
 	github.com/containernetworking/cni v1.1.2
 	github.com/containernetworking/plugins v1.2.0
-	github.com/containers/common v0.51.0
+	github.com/containers/common v0.51.1
 	github.com/containers/image/v5 v5.24.2
 	github.com/cyphar/filepath-securejoin v0.2.3
 	github.com/docker/distribution v2.8.1+incompatible
--- a/go.sum
+++ b/go.sum
@@ -121,8 +121,8 @@ github.com/containernetworking/cni v1.1.2 h1:wtRGZVv7olUHMOqouPpn3cXJWpJgM6+EUl3
 github.com/containernetworking/cni v1.1.2/go.mod h1:sDpYKmGVENF3s6uvMvGgldDWeG8dMxakj/u+i9ht9vw=
 github.com/containernetworking/plugins v1.2.0 h1:SWgg3dQG1yzUo4d9iD8cwSVh1VqI+bP7mkPDoSfP9VU=
 github.com/containernetworking/plugins v1.2.0/go.mod h1:/VjX4uHecW5vVimFa1wkG4s+r/s9qIfPdqlLF4TW8c4=
-github.com/containers/common v0.51.0 h1:Ax4YHNTG8cEPHZJcMYRoP7sfBgOISceeyOvmZzmSucg=
-github.com/containers/common v0.51.0/go.mod h1:3W2WIdalgQfrsX/T5tjX+6CxgT3ThJVN2G9sNuFjuCM=
+github.com/containers/common v0.51.1 h1:xMQpFSuUovZgPRWJu1+uBixKsPcRrD4aHADkouz2K9Q=
+github.com/containers/common v0.51.1/go.mod h1:3W2WIdalgQfrsX/T5tjX+6CxgT3ThJVN2G9sNuFjuCM=
 github.com/containers/image/v5 v5.24.2 h1:QcMsHBAXBPPnVYo6iEFarvaIpym7sBlwsGHPJlucxN0=
 github.com/containers/image/v5 v5.24.2/go.mod h1:oss5F6ssGQz8ZtC79oY+fuzYA3m3zBek9tq9gmhuvHc=
 github.com/containers/libtrust v0.0.0-20230121012942-c1716e8a8d01 h1:Qzk5C6cYglewc+UyGf6lc8Mj2UaPTHy/iF2De0/77CA=


cmd/internal/cli/action_flags.go
```diff
@@ -812,6 +813,15 @@ var actionOCIFlag = cmdline.Flag{
 	EnvKeys:      []string{"OCI"},
 }
 
+// -B|--bind
+var actionCDIDevices = cmdline.Flag{
+	ID:           "actionCDIDevices",
+	Value:        &cdiDevices,
+	DefaultValue: []string{},
+	Name:         "devices",
+	Usage:        "a CDI device mapping. A device mapping consists of a kind specification followed by the equal sign (=) and a device label (e.g. vendor.com/device=mydevice). Kind specifications consists of a prefix and a name, separated by a slash (e.g. vendor.com/device, in the previous example). Multiple device mappings can be given by a comma separated list.",
```
dtrudg, line 822:
```suggestion
	Usage:        "a CDI device mapping. A device mapping consists of a kind specification followed by the equal sign (=) and a device label (e.g. vendor.com/device=mydevice). Kind specifications consist of a prefix and a name, separated by a slash (e.g. vendor.com/device, in the previous example). Multiple device mappings can be given by a comma separated list.",
```

internal/pkg/runtime/launcher/oci/launcher_linux.go
```diff
@@ -219,7 +219,15 @@ func (l *Launcher) createSpec() (*specs.Spec, error) {
 	if err != nil {
 		return nil, err
 	}
-	spec.Mounts = mounts
+	spec.Mounts = append(spec.Mounts, mounts...)
+
+	for _, cdiDevice := range l.cfg.CDIDevices {
```
dtrudg, line 224: Although this is a small and simple loop, please could you pull it out into a function so it ends up as `getCDIMounts` returning a slice to append. Ideally in a `cdi_linux.go` file. It's just a preference for separating stuff a bit more as this launcher code is going to see a lot of work, and expansion for 4.0.

internal/pkg/runtime/launcher/oci/spec_linux.go
```diff
@@ -115,3 +118,25 @@ func addNamespaces(spec specs.Spec, ns launcher.Namespaces) specs.Spec {
 
 	return spec
 }
+
+// addCDIDevice adds the requested device to an existing spec.
+func addCDIDevice(spec specs.Spec, cdiDevice string) (specs.Spec, error) {
```
dtrudg, line 123: Please could you move this into a new `cdi_linux.go` file here. It's just a preference for separating stuff a bit more as this launcher code is going to see a lot of work, and expansion for 4.0.

internal/pkg/runtime/launcher/util.go
```diff
@@ -46,3 +48,8 @@ func HidepidProc() bool {
 	}
 	return false
 }
+
+// IsCDIDevice checks whether a string is a valid CDI device selector.
+func IsCDIDevice(str string) bool {
```
dtrudg, line 53: Let's bring this across to `cdi_linux.go` too. The other things here should end up getting factored out into a higher level utils pkg by 4.0.

internal/pkg/runtime/launcher/options.go
```diff
@@ -141,6 +141,9 @@ type Options struct {
 	// SysContext holds Docker/OCI image handling configuration.
 	// This will be used by a launcher handling OCI images directly.
 	SysContext *types.SystemContext
+
+	// CDIDevices contains the list of CDI device mappings (if any)
+	CDIDevices []string
```
dtrudg, line 146: As at the CLI layer, lets call this `Device` in case (a) we end up implementing non-CDI `--device` binding, and (b) it moves out of `internal` for 4.0... which is an aim... sensible semver controlled public API.

internal/pkg/runtime/launcher/oci/launcher_linux.go
dtrudg: Actually - I see why the above isn't easy, sorry.... there's only a path to mutate the spec.

Let's go for an `addCDIDevices` function, then... which will take the spec and modify it.

dtrudg: Yeah - per my follow-up, ignore that. I had thought it might be easier to get mounts, env vars etc out individually.. but I should have checked in the CDI code first.

internal/pkg/runtime/launcher/options.go
```diff
@@ -141,6 +141,9 @@ type Options struct {
 	// SysContext holds Docker/OCI image handling configuration.
 	// This will be used by a launcher handling OCI images directly.
 	SysContext *types.SystemContext
+
+	// Device contains the list of device mappings (if any), e.g. CDI mappings.
+	Device []string
```
elezar, line 146: nit: Should this be `Devices` since it is a slice?

internal/pkg/runtime/launcher/oci/cdi_linux.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+
+	"github.com/container-orchestrated-devices/container-device-interface/pkg/cdi"
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+// addCDIDevice adds an array of CDI devices to an existing spec.
+func addCDIDevice(spec *specs.Spec, cdiDevices []string) error {
```
elezar, line 19: nit: `addCDIDevices` since this adds multiple devices?

internal/pkg/runtime/launcher/oci/cdi_linux.go
```diff
@@ -0,0 +1,53 @@
...
+	// Get the CDI registry, passing a cdi.WithAutoRefresh(false) option so that CDI registry files are not scanned asynchronously. (We are about to call a manual refresh, below.)
+	registry := cdi.GetRegistry(cdi.WithAutoRefresh(false))
+
+	// Refresh the CDI registry.
+	if err := registry.Refresh(); err != nil {
+		return fmt.Errorf("Error encountered refreshing the CDI registry: %v", err)
+	}
+
+	for _, cdiDevice := range cdiDevices {
```
elezar, line 28: Note that the `registry.InjectDevices` function accepts a variadic list of fully-qualified device names. This includes a check for a valid name. As such I don't know whether introducing an `addSingleCDIDevice` function adds specific value.

internal/pkg/runtime/launcher/options.go
```diff
@@ -490,3 +493,11 @@ func OptSysContext(sc *types.SystemContext) Option {
 		return nil
 	}
 }
+
+// OptDevice sets CDI device mappings to apply.
+func OptDevice(op []string) Option {
+	return func(lo *Options) error {
+		lo.Device = op
```
elezar, line 500: Does it make sense to call `IsQualifiedName` for each of the elements of `op` here to validate the options. Then it should not be required to perform this check explicitly later and the check for further validity can be delegated to the registry.

cmd/internal/cli/action_flags.go
```diff
@@ -812,6 +813,15 @@ var actionOCIFlag = cmdline.Flag{
 	EnvKeys:      []string{"OCI"},
 }
 
+// -B|--bind
+var actionDevice = cmdline.Flag{
+	ID:           "actionDevice",
+	Value:        &device,
+	DefaultValue: []string{},
+	Name:         "device",
+	Usage:        "a CDI device mapping. A device mapping consists of a kind specification followed by the equal sign (=) and a device label (e.g. vendor.com/device=mydevice). Kind specifications consist of a prefix and a name, separated by a slash (e.g. vendor.com/device, in the previous example). Multiple device mappings can be given by a comma separated list.",
```
elezar, line 822: I think the terms we use are `fully-qualified CDI device name` instead of `mapping`. The parts are:

`<VENDOR>/<CLASS>=<NAME>` and together these are a "fully-qualified CDI device name".

cmd/internal/cli/action_flags.go
```diff
@@ -812,6 +813,15 @@ var actionOCIFlag = cmdline.Flag{
 	EnvKeys:      []string{"OCI"},
 }
 
+// -B|--bind
```
elezar, line 816: This comment seems incorrect.

cmd/internal/cli/action_flags.go
```diff
@@ -37,6 +37,7 @@ var (
 	singularityEnvFile string
 	noMount            []string
 	proot              string
+	device             []string
```
elezar, line 40: `devices`?

elezar (commented):
Some comments from the perspective of CDI. Here I would obviously defer to singulartiy maintainers.

One thing that I was wonder was whether we want to make the use of CDI devices mutually exclusive with other mechanisms for injecting NVIDIA devices? I would argue that this would make sense sicne using them both may lead to hard-to-diagnose problems due to the interactions with the `nvidia-container-runtime-hook`. This could be implemented as a follow-up though.

internal/pkg/runtime/launcher/options.go
dtrudg: Not really... this could be open to non-CDI `--device` usage in future, and as there's an intention to move the `Opt*` and stuff out of `internal`, for public consumption... I'd rather keep CDI specific stuff separate at present.

cmd/internal/cli/action_flags.go
dtrudg: Definitely just `device` here at present please. We are matching the existing pattern for other flag vars where the var is named the same as the flag, which is non-plural `--device`.

If we change it to plural (for which I see the argument, and tend to agree), we'll do that consistently across our other flag vars too, in a separate re-factoring.

dtrudg:
> One thing that I was wonder was whether we want to make the use of CDI devices mutually exclusive with other mechanisms for injecting NVIDIA devices? I would argue that this would make sense sicne using them both may lead to hard-to-diagnose problems due to the interactions with the `nvidia-container-runtime-hook`.

I don't think we want to do this, personally... other than perhaps issuing a warning. I'd only support making them mutually exclusive if there's a way to scope it easily to *only* NVIDIA CDI devices.

I can definitely foresee situations where someone might be on a system with no NVIDIA CDI configuration in place yet, but some other CDI configuration for another device. We see the most unlikely combinations of config on hosts, quite regularly :-)

elezar:
> > One thing that I was wonder was whether we want to make the use of CDI devices mutually exclusive with other mechanisms for injecting NVIDIA devices? I would argue that this would make sense sicne using them both may lead to hard-to-diagnose problems due to the interactions with the `nvidia-container-runtime-hook`.
> 
> I don't think we want to do this, personally... other than perhaps issuing a warning. I'd only support making them mutually exclusive if there's a way to scope it easily to _only_ NVIDIA CDI devices.
> 
> I can definitely foresee situations where someone might be on a system with no NVIDIA CDI configuration in place yet, but some other CDI configuration for another device. We see the most unlikely combinations of config on hosts, quite regularly :-)

You're right. I forgot about the more general use cases here. Sorry about that.

dtrudg (approved):
This looks good to me to merge as the initial CDI implementation.

If there are any additional things you want to do, please ask @wobito or another to re-review while I'm travelling.

We wil handle a bit more e2e-testing later on. There are the GPU tests, in which we can do some stuff to actually test whether an Nvidia GPU is visible in the container... on systems with one of those :-)

PR #1495, opened 2023-03-29
oci: support --hostname

### LINKED ISSUE #1474: Support `--hostname` in `--oci` mode

preminger:

Support `--hostname` in OCI mode. Requires UTS namespace, so in OCI mode, `--hostname` will infer `--uts`. Added test in e2e-test suite (e2e/actions/oci.go:148) for this functionality.

 - Fixes #1474

### DIFF
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -52,7 +52,7 @@ func getCacheHandle(cfg cache.Config) *cache.Handle {
 //   - do the proper path unsetting;
 //   - and implement flag inferences for:
 //     --compat
-//     --hostname (OCI-mode only)
+//     --hostname
 func actionPreRun(cmd *cobra.Command, args []string) {
 	// For compatibility - we still set USER_PATH so it will be visible in the
 	// container, and can be used there if needed. USER_PATH is not used by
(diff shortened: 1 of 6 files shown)


cmd/internal/cli/actions.go
```diff
@@ -67,6 +72,11 @@ func actionPreRun(cmd *cobra.Command, args []string) {
 		noUmask = true
 		noEval = true
 	}
+
+	// --hostname in OCI mode requires UTS namespace
+	if ociRuntime && (len(hostname) > 0) {
```
dtrudg, line 77: `--hostname` for non OCI mode will also require a UTS namespace. Are you able to find where it is adding that? 

It'd be good to either...

* Do it at the same level for `--oci`, so it's consistent. -or-
* Do it for the OCI and non-OCI mode in the same place, here.

dtrudg: Probably in native/launcher.go?

dtrudg (commented):
LGTM - with the suggestion it'd be nice to ensure we are handling the `--hostname` inferring `--uts` in a consistent way for non-oci mode too... and in one place if that's possible.

PR #1498, opened 2023-03-30
oci: support --scratch

### LINKED ISSUE #1482: Support `--scratch` in `--oci` mode

preminger:

Support `--scratch` (shorthand: `-S`) for mounting a tmpfs scratch directory in the container in oci mode, just like we do in native mode.

### Testing notes

Added e2e tests for this functionality in oci mode, mirroring and extending the e2e tests for the corresponding native-mode functionality. Added a couple of tests to both modes:

- Test that a scratch-mounted dir is indeed present in the container as a mount with `tmpfs` filesystem
- Test that when a bind-mounted dir shares a path with a scratch-mounted dir, the bind-mounted one will override the scratch-mounted one. (This was already behavior in native mode, though it wasn't being tested, as far as I could tell. Now it's enforced by the e2e tests for each of the two modes.)

 - Fixes #1482

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,9 @@
 - The `remote status` command will now print the username, realname, and email
   of the logged-in user, if available.
 - OCI mode now supports `--hostname` (requires UTS namespace, therefore this
-  flag will infer `--uts` if running in OCI mode).
+  flag will infer `--uts`).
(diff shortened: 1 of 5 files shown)


dtrudg (approved):
LGTM. Thanks!

PR #1523, opened 2023-04-05
oci: fix custom cwd not respected when homedir set

preminger:

There was a problem where, in OCI mode, a custom working directory (`--cwd`/`--pwd`) would not be respected, and the homedir would be used as the working directory instead.

Interestingly, this was not caught by our e2e tests, which (ultimately) revealed a problem in the way I was handling profiles in the `actionOciExec()` function (e2e/actions/oci.go:228).

This fixes both issues.

### DIFF
--- a/e2e/actions/oci.go
+++ b/e2e/actions/oci.go
@@ -120,10 +120,11 @@ func (c actionTests) actionOciExec(t *testing.T) {
 	homePath := filepath.Join("/home", basename)
 
 	tests := []struct {
-		name        string
-		argv        []string
-		exit        int
-		wantOutputs []e2e.SingularityCmdResultOp
+		name         string
+		argv         []string
+		exit         int
+		wantOutputs  []e2e.SingularityCmdResultOp
+		skipProfiles map[string]bool
 	}{
 		{
 			name: "NoCommand",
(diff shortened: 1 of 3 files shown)


internal/pkg/runtime/launcher/oci/process_linux.go
```diff
@@ -88,6 +88,10 @@ func getProcessArgs(imageSpec imgspecv1.Image, process string, args []string) []
 // Currently this is the user's tmpfs home directory (see --containall).
 // Because this is called after mounts have already been computed, we can count on l.cfg.HomeDir containing the right value, incorporating any custom home dir overrides (i.e., --home).
 func (l *Launcher) getProcessCwd() (dir string, err error) {
+	if len(l.cfg.CwdPath) > 1 {
```
dtrudg, line 91: This needs to be `>0` or `--cwd /` won't work...

```
09:23 am $ singularity run --oci --cwd / --home /bob docker://alpine pwd
/bob

09:23 am $ singularity run --oci --cwd /. --home /bob docker://alpine pwd
/
```

e2e/actions/oci.go
```diff
@@ -172,6 +173,10 @@ func (c actionTests) actionOciExec(t *testing.T) {
 				e2e.ExpectOutput(e2e.RegexMatch, `\bHOME=/myhomeloc\b`),
 				e2e.ExpectOutput(e2e.RegexMatch, `\btmpfs on /myhomeloc\b`),
 			},
+			skipProfiles: map[string]bool{
```
dtrudg, line 176: These should definitely work, but it's a separate issue from the problem that's fixed here, so we'll handle it separately.

dtrudg (changes requested):

dtrudg (approved):

Issue #1529, opened 2023-04-06
oci: --home not working with --fakeroot or sudo


preminger:
**Version of Singularity**

`main` development branch

**Describe the bug**

In OCI mode (`--oci`), the `--home` flag works as intended when running `singularity` as a regular user, but not when running with `--fakeroot` or when running as root on the host machine.

**To Reproduce**

```
singularity shell --oci --fakeroot --home /myhomedir docker://alpine
2023/04/06 12:25:44  info unpack layer: sha256:c41833b44d910632b415cd89a9cdaa4d62c9725dc56c99a7ddadafd6719960f9
Singularity> cd
Singularity> pwd
/root
Singularity>
```

```
sudo singularity shell --oci --home /myhomedir docker://alpine
2023/04/06 12:26:14  info unpack layer: sha256:c41833b44d910632b415cd89a9cdaa4d62c9725dc56c99a7ddadafd6719960f9
Singularity> cd
Singularity> pwd
/root
Singularity>
```

**Expected behavior**

Should behave like regular user-run `singularity`:

```
sudo singularity shell --oci --home /myhomedir docker://alpine
2023/04/06 12:26:14  info unpack layer: sha256:c41833b44d910632b415cd89a9cdaa4d62c9725dc56c99a7ddadafd6719960f9
Singularity> cd
Singularity> pwd
/root
Singularity>
```

**OS / Linux Distribution**

```
$ cat /etc/os-release
NAME="Fedora Linux"
VERSION="37 (KDE Plasma)"
ID=fedora
VERSION_ID=37
VERSION_CODENAME=""
PLATFORM_ID="platform:f37"
PRETTY_NAME="Fedora Linux 37 (KDE Plasma)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:37"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f37/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=37
REDHAT_SUPPORT_PRODUCT="Fedora"
[...]

PR #1531, opened 2023-04-06
revamped Passwd() utility-function

preminger:

The way the Passwd() utility-function (internal/pkg/util/fs/files/passwd.go:22) worked until now, it was only able to append a new line to what it got as input (from the base layer).

This meant that, for example, to override the homedir of a regular user, we would simply add a modified line (with the custom homedir) to the in-container /etc/passwd; but to override the homedir of root (when running singularity itself as root, or when running with `--fakeroot`), we couldn't do that – since there was already a "root" line in the base layer's /etc/passwd – and we resorted to an entirely different method, namely, overriding the `$HOME` environment variable upon container initialization.

In addition to being non-uniform, the approach involving environment variables can also give rise to an internally inconsistent state, where, when we're root inside the container, our $HOME is set to one thing (a custom homedir) but the /etc/passwd file still has `/root` as the homedir for the root user.

A more uniform approach, which also avoids this inconsistent state described above, is possible: change the Passwd() utility-function so that when "adding" a line to /etc/passwd it first checks if a line with the corresponding uid already exists, and if it does, replaces that specific line instead of appending a new line to the file.

### Performance considerations

The implementation in this PR involves calling `ParsePasswdLine()` from the [github.com/stat0s2p/etcpwdparse](https://pkg.go.dev/github.com/stat0s2p/etcpwdparse) package on every line in the /etc/passwd file (internal/pkg/util/fs/files/passwd.go:52).

If this turns out to be problematic from a performance perspective, we can replace it with strings.split or something similar to just pull out the first field of each /etc/passwd line (=the username), and using only that to judge sameness-of-users for the purposes of deciding whether to replace a line or append a new one.

### Related PRs

Bug #1529 was a failure to override root's homedir when running in OCI mode as root/fakeroot. It was fixed in PR #1530 using the environment-variables method described above, in order to maintain uniformity with the native-runtime mode, and in order to not have the fix depend on the redoing of Passwd() described here.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,7 @@ require (
 	github.com/samber/lo v1.38.1
 	github.com/seccomp/libseccomp-golang v0.10.0
 	github.com/shopspring/decimal v1.3.1
-	github.com/sigstore/sigstore v1.6.2
+	github.com/sigstore/sigstore v1.6.3
 	github.com/sirupsen/logrus v1.9.0
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
@@ -178,7 +178,7 @@ require (
 	golang.org/x/net v0.9.0 // indirect
 	golang.org/x/sync v0.1.0 // indirect
 	golang.org/x/tools v0.7.0 // indirect
-	google.golang.org/genproto v0.0.0-20230331144136-dcfb400f0633 // indirect
+	google.golang.org/genproto v0.0.0-20230410155749-daa745c078e1 // indirect
 	google.golang.org/grpc v1.54.0 // indirect
 	google.golang.org/protobuf v1.30.0 // indirect
 	gopkg.in/go-jose/go-jose.v2 v2.6.1 // indirect
--- a/go.sum
+++ b/go.sum
@@ -500,8 +500,8 @@ github.com/sigstore/fulcio v1.2.0 h1:I4H764cDbryKXkPtasUvo8bcix/7xLvkxWYWNp+JtWI
 github.com/sigstore/fulcio v1.2.0/go.mod h1:FS7qpBvOEqs0uEh1+hJxzxtJistWN29ybLtAzFNUi0c=
 github.com/sigstore/rekor v1.1.0 h1:9fjPvW0WERE7VPtSSVSTbDLLOsrNx3RtiIeZ4/1tmDI=
 github.com/sigstore/rekor v1.1.0/go.mod h1:jEOGDGPMURBt9WR50N0rO7X8GZzLE3UQT+ln6BKJ/m0=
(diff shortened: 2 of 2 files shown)


preminger:
PLEASE NOTE THE "PERFORMANCE CONSIDERATIONS" SECTION IN THE INITIAL COMMENT.

dtrudg (changes requested):
A CHANGELOG.md entry detailing the behaviour change is required (w.r.t. what happens with /etc/passwd vs $HOME before and after this change).

General approach is fine, but the dependencies added here need to be reviewed / replaced.

https://github.com/stat0s2p/etcpwdparse is a fork of https://github.com/astromechza/etcpwdparse without any changes. In this type of situation, generally prefer the original. Also, that repo doesn't have a `go.mod`. If there's an alternative that is using modern go modules it'd be preferred.

github.com/revel/cmd probably isn't a good choice to import just to access a function to read all lines from a file. revel is a large web framework, and revel/cmd the CLI stuff for it, so it's not surprising it brings in quite a lot of other dependencies. These will (such is the nature of a web framework) introduce quite a lot of dependency churn vs the benefit.

Using bufio.Scanlines (i.e. this pattern https://www.golangprograms.com/golang-read-file-line-by-line-to-string.html) is more verbose, but will avoid bringing in 7 dependencies.

preminger:
Concerning the last part: if I understood your question correctly, you're talking about a situation where the base layer for the singularity container already brings in a `/etc/passwd` file containing a user `dave`, and you're also running singularity in user mode (i.e., not as root and not with `--fakeroot`) from a host user called `dave`.

Looking at the code *before* this change, it looks like what would have happened is that we would write an additional line for `dave` into the container's `/etc/passwd` anyway (internal/pkg/util/fs/files/passwd.go:40-44). And, based on how it behaved when there were two lines for the user `root`, I have to assume it would simply ignore the second `dave` line, and so this additional line would have no effect.

The new code, as it stands, would *replace* the existing `dave` line in the container's `/etc/passwd`, **but only if both `dave`s had the same user id**. In fact, it would replace the `dave` line in the container's `/etc/passwd` with a line for `omer` if we were running as `omer` in the host, and `omer`'s user id in the host was the same as `dave`'s user id in the container's base layer.

I can see how this might introduce fairly significant backward-compatibility issues, depending on which use cases are out there.

Please advise.

preminger:
Oh, sorry @dtrudg – I think I saw a previous version of your last comment that said something like "let's mirror the behavior that Docker has" and I was contemplating the best chance to find out what that was (spelunking in the docs, or just poking around with docker myself), but I see now that you've edited the comment. I think I'm good to go from here. Thanks!

dtrudg (commented):
Looks sensible to me. Please could you rebase against main, and I'll take a final look before I put a :heavy_check_mark:  on it.

PR #1551, opened 2023-04-12
oci: support resolv.conf and --dns

### LINKED ISSUE #1472: Support `resolv.conf` and `--dns` in `--oci` mode
In the native runtime, an `/etc/resolv.conf` file is prepared and mounted in to the container by default. If the `--dns` flag is not provided then the content of this file is the same as `/etc/resolv.conf` on the host.

The `--dns` flag allows custom dns servers to be set for the container. If specified, the `/etc/resolv.conf` that is bound into the container will be overridden to contain only `nameserver` entries for the specified DNS servers.

The experimental `--oci` mode currently has no resolv.conf handling, so DNS setup has to been done manually. The `--oci` mode should implement the same approach as the native runtime.

The implementation in the native runtime can be used for guidance. Note the utility function `files.ResolvConf` that is available.

https://github.com/sylabs/singularity/blob/7d8e39cae31b33119bc6660c7b8e7f7f522fc40b/internal/pkg/runtime/engine/singularity/container_linux.go#L2209

preminger:

Brought handling of `resolv.conf` and `--dns` flag in OCI mode on par with their behavior in native mode.

In particular: an `/etc/resolv.conf` file is always created in the OCI container. If a `--dns` argument is provided, its contents (a comma-separated list of IPs of nameservers) determines the content of this `resolv.conf` file. Otherwise, the content is taken from the host's `/etc/resolv.conf` file.

 - Fixes #1472


preminger:
Unless I've missed something, there's nothing in the current e2e suite that tests `--dns`, or this `resolv.conf` functionality more generally, not even in native mode... If this is indeed the case, perhaps we should open a separate github issue to write e2e tests for this functionality in both modes?

preminger:
For now, I've added some e2e tests here for both native- and OCI-mode `--dns`. There are two tests for each mode: one with google's DNS servers, and one with cloudflare's. The idea here is that you want to _really_ test the `--dns` functionality, and there's a chance the CI environment already has its DNS servers set to, e.g., google's, in which case that particular test wouldn't truly be testing the functionality of overriding the host's DNS servers.

internal/pkg/runtime/launcher/oci/launcher_linux.go
```diff
@@ -347,6 +351,43 @@ func (l *Launcher) updatePasswdGroup(rootfs string, uid, gid uint32) error {
...
+	} else {
+		resolvConfData, err = os.ReadFile(hostResolvConfPath)
+		if err != nil {
+			return fmt.Errorf("could not read host's resolv.conf file: %w", err)
+		}
+	}
+
+	stat, err := os.Stat(containerEtc)
+	if os.IsNotExist(err) || !stat.IsDir() {
+		sylog.Warningf("container does not contain an /etc directory; skipping resolve.conf configuration")
```
dtrudg, line 380:
```suggestion
		sylog.Warningf("container does not contain an /etc directory; skipping resolv.conf configuration")
```

dtrudg (changes requested):
This looks good, test approach seems fair to me.

The only additional thing we need to do is make sure that resolv.conf is left alone if `config resolv_conf = no` in `singularity.conf`, i.e. `l.singularityConf.ConfigResolvConf` is false in the launcher code.

The majority of manipulations of files or binds are gated by a setting in `singularity.conf`. It's worth a look in there to check whether there is a relevant option to consider. Apologies for not raising this in the issue.

dtrudg:
Actually.... I'm going to change track here. Let's accept this as is, and I'll add the code that honours the singularity.conf directive to my PR for those.... makes sense to keep together with the relevant config tests.

dtrudg:
Added the singularity.conf directive conditional to https://github.com/sylabs/singularity/pull/1554 - where all the others are addressed.

PR #1539, opened 2023-04-11
fix: oci: enter cgroup before executing crun as non-root

dtrudg:
## Description of the Pull Request (PR):

When executed from a root-owned cgroup, such as the session scope resulting from a bare ssh login, crun will fail to create our requested container cgroup.

If we are running as non-root, create and move into a user-owned cgroup, so that there's a common user-owned ancestor. This avoids the `crun` error.

Note that no workaround is needed for `runc` as it is able to create the requested container cgroup without any issue.

### This fixes or addresses the following GitHub issues:

 - Fixes #1538 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/main/CONTRIBUTORS.md)

PR #1570, opened 2023-04-17
fix: oci: Don't create cgroup for crun on v1 / cgroupfs

### LINKED ISSUE #1569: crun --oci cgroup nesting causes error on cgroups  v1 as non-root
**Version of Singularity**

main

**Describe the bug**

After #1539 attempting to run a container in `--oci` mode on a cgroups v1 system using `crun` will fail with an error. This can be seen in the e2e-tests on RHEL8...

```
    === CONT  TestE2E/PAR/DOCKER/ordered/user/OCIUser
        docker.go:919:
            "/usr/bin/singularity run --oci docker://sylabsio/docker-user" output:
            FATAL:   while applying crun cgroup workaround: couldn't create cgroup manager: rootless cgroups requires cgroups v2
```

**To Reproduce**

`singularity run --oci docker://sylabsio/docker-user` on a system using cgroups v1 and crun (not runc).

**Expected Behavior**

Under cgroups v1 as a normal user, we cannot create a cgroup to move into before calling `crun`,  so we should not try to do this.

In this mode of operation (cgroups v1, unprivileged), `crun` won't be trying to do the same cgroup manipulation as on a v2 system. Therefore we (a) generally don't hit the issue #1539 works around... and (b) have no ability to rectify any cgroup constraints that will cause a different failure.

dtrudg:

If we are running under cgroups v1 or with the cgroupfs manager (i.e. not systemd as cgroup manager), do not attempt to enter a cgroup at startup with crun. We cannot create a cgroup unprivileged in this situation.

Under cgroups v1, crun will not perform the cgroups manipulation that leads to the issue we worked around in #1539. Any other issue with the cgroup that we are in at launch cannot be rectified, either.

 - Fixes #1569 

### DIFF
--- a/internal/pkg/runtime/launcher/oci/launcher_linux.go
+++ b/internal/pkg/runtime/launcher/oci/launcher_linux.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/container-orchestrated-devices/container-device-interface/pkg/cdi"
 	"github.com/google/uuid"
+	lccgroups "github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sylabs/singularity/internal/pkg/buildcfg"
 	"github.com/sylabs/singularity/internal/pkg/cache"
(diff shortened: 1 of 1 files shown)


preminger (approved):
LGTM!

PR #1594, opened 2023-04-24
oci: fix to not override HOME when container specifies USER (+ e2e test)

### LINKED ISSUE #1593: bug: HOME is overridden in OCI mode even if container specifies USER
As part of PR #1530, we introduced a bug where, in OCI mode, if the container specifies its own user, we'd still overwrite the `HOME` environment variable inside the container.

**To Reproduce**

```
singularity shell --oci docker://sylabsio/docker-user
2023/04/24 13:42:57  info unpack layer: 
...
Singularity> echo $HOME
/home/omer
Singularity> cd; pwd
/home/omer
Singularity>
```

**Expected behavior**

```
singularity shell --oci docker://sylabsio/docker-user
2023/04/24 13:42:57  info unpack layer: 
...
Singularity> echo $HOME
/home/testuser
Singularity> cd; pwd
/home/testuser
Singularity>
```

preminger:

In OCI mode, avoid overriding container's HOME environment variable if container specifies its own USER.

Also add e2e test of the above.

 - Fixes #1593

### DIFF
--- a/e2e/docker/docker.go
+++ b/e2e/docker/docker.go
@@ -883,35 +883,86 @@ func (c ctx) testDockerCMDQuotes(t *testing.T) {
 
 // Check that the USER in a docker container is honored under --oci mode
 func (c ctx) testDockerUSER(t *testing.T) {
+	dockerURI := "docker://sylabsio/docker-user"
 	tests := []struct {
-		name         string
-		expectOutput string
-		profile      e2e.Profile
+		name          string
+		cmd           string
+		args          []string
+		expectOutputs []e2e.SingularityCmdResultOp
+		profile       e2e.Profile
+		expectExit    int
 	}{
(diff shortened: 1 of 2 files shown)


dtrudg:
> Should the added e2e tests here (currently in e2e/docker/docker.go:936-966) go in a separate function called `issue1593()` or something of that sort, or is it okay to leave them where they are? The advantage of leaving them where they are, imo, is that the `testDockerUSER()` function _is_ where they belong, thematically speaking.

It's best that the go in testDockerUSER, where you put them. Generally the `issueXXX` tests are used for testing for regressions, where something doesn't neatly fit a more general test, or we want to be especially explicit about testing for a particular regression for some reason.

dtrudg (approved):
LGTM. Thanks!

PR #1597, opened 2023-04-25
Revamp passwd utility function

preminger:

_NOTE: This PR is a continuation of #1531. See additional discussion there._

The way the Passwd() utility-function (internal/pkg/util/fs/files/passwd.go:22) worked until now, it was only able to append a new line to what it got as input (from the base layer).

This meant that, for example, to override the homedir of a regular user, we would simply add a modified line (with the custom homedir) to the in-container /etc/passwd; but to override the homedir of root (when running singularity itself as root, or when running with `--fakeroot`), we couldn't do that – since there was already a "root" line in the base layer's /etc/passwd – and we resorted to an entirely different method, namely, overriding the `$HOME` environment variable upon container initialization.

In addition to being non-uniform, the approach involving environment variables can also give rise to an internally inconsistent state, where, when we're root inside the container, our $HOME is set to one thing (a custom homedir) but the /etc/passwd file still has `/root` as the homedir for the root user.

A more uniform approach, which also avoids this inconsistent state described above, is possible: change the Passwd() utility-function so that when "adding" a line to /etc/passwd it first checks if a line with the corresponding uid already exists, and if it does, replaces that specific line instead of appending a new line to the file.

### Performance considerations

The implementation in this PR involves calling `ParsePasswdLine()` from the [github.com/stat0s2p/etcpwdparse](https://pkg.go.dev/github.com/stat0s2p/etcpwdparse) package on every line in the /etc/passwd file (internal/pkg/util/fs/files/passwd.go:52).

If this turns out to be problematic from a performance perspective, we can replace it with strings.split or something similar to just pull out the first field of each /etc/passwd line (=the username), and using only that to judge sameness-of-users for the purposes of deciding whether to replace a line or append a new one.

### Related PRs

Bug #1529 was a failure to override root's homedir when running in OCI mode as root/fakeroot. It was fixed in PR #1530 using the environment-variables method described above, in order to maintain uniformity with the native-runtime mode, and in order to not have the fix depend on the redoing of Passwd() described here.


internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -2139,9 +2139,7 @@ func (c *container) addFilesMount(system *mount.System) error {
 }
 
 func (c *container) addIdentityMount(system *mount.System) error {
-	if (os.Geteuid() == 0 && c.engine.EngineConfig.GetTargetUID() == 0) ||
-		c.engine.EngineConfig.GetFakeroot() {
-		sylog.Verbosef("Not updating passwd/group files, running as root!")
+	if c.engine.EngineConfig.GetNoHome() {
```
dtrudg, line 2142: Sorry - on a second pass over it I don't think that this is what we want.

`--no-home` is just about not mounting the HOME directory into the container.

It's actually (confusingly) valid to do something like `--no-home --home /opt`... in which case HOME should be `/opt` in the container... but not  `/opt` will not be mounted from outside.

Also, if I'm user `dave`, then even if I run with `--no-home` then we need to make sure `dave` is present in the container `/etc/passwd`.

Apologies for not picking this up before. I will need to find a couple of hours to think about this and trawl through some old discussions from a few years ago to satisfy myself what the right approach is.

dtrudg (commented):
It LGTM apart from what I've mentioned in the native runtime code. That's on me... I need to think about it for a while. Apologies.

internal/pkg/runtime/engine/singularity/container_linux.go
dtrudg: Right... having combed through the code and the history of this in the morning... it looks to me like we should just always update the passwd/group file.

Let's do that... we're swapping 'don't update only if root' for 'always update'. On a Linux host you can certainly change root's home directory from `/root` if you want to. It's not necessarily a sensible thing, but you can do it... so `--home` should do it too, for consistency.

If I've missed anything... we've got a few months for it to surface before this goes out in a 4.0 :-)

dtrudg (dismissed):

dtrudg (approved):

PR #1659, opened 2023-05-08
oci: support --overlay

### LINKED ISSUE #1478: Support `--overlay` in `--oci` mode (directory)
When `--overlay <directory>` is specified on an `--oci` `run/exec/shell` command, then _directory_ should be mounted as an overlay on top of the rootfs of the container. This will allow any changes in the container filesystem to be persisted, and re-used across different runs.

:crossed_fingers:  this should be relatively straightforward. After #1622 we are, by default, setting up a `tmpfs` as an ephemeral overlay on to of the container rootfs. The `--overlay` flag needs to disable that tmpfs overlay, and instead perform a directory based overlay.

Most of the work required is likely to be around passing the details of the overlay through to the `singularity oci run-wrapped` command that is called by the oci launcher.

Adding a command line flag(s) onto `oci run-wrapped` is probably an appropriate and simple method here. There might be a need to start re-factoring some function signatures, though... so they don't grow to taking very large numbers of parameters.

Ref (native mode): https://docs.sylabs.io/guides/3.11/user-guide/persistent_overlays.html#directory-overlay

preminger:

Support `--overlay <overlay_dir>` flag in OCI-mode (in "action" commands, i.e., run/exec/shell). This allows the user to mount a directory as a filesystem overlay, in lieu of using a writable tmpfs (which is the default behavior since version 3.11.3). This enables writes to the filesystem to persist across runs of the OCI container. If `<overlay_dir>` does not exist, Singularity will attempt to create it.

 - Fixes #1478

### DIFF
--- a/pkg/ocibundle/tools/overlay_linux.go
+++ b/pkg/ocibundle/tools/overlay_linux.go
@@ -12,40 +12,58 @@ import (
 	"syscall"
 )
 
-// CreateOverlay creates a writable overlay based on a directory.
+// CreateOverlay creates a writable overlay based on a directory inside the OCI bundle.
 func CreateOverlay(bundlePath string) error {
-	var err error
-
 	oldumask := syscall.Umask(0)
 	defer syscall.Umask(oldumask)
 
-	overlayDir := filepath.Join(bundlePath, "overlay")
-	if err = os.Mkdir(overlayDir, 0o700); err != nil {
-		return fmt.Errorf("failed to create %s: %s", overlayDir, err)
+	overlayPath := filepath.Join(bundlePath, "overlay")
+	var err error
+	if err = os.Mkdir(overlayPath, 0o700); err != nil {
+		return fmt.Errorf("failed to create %s: %s", overlayPath, err)
 	}
 	// delete overlay directory in case of error
 	defer func() {
 		if err != nil {
-			os.RemoveAll(overlayDir)
+			os.RemoveAll(overlayPath)
 		}
 	}()
 
-	err = syscall.Mount(overlayDir, overlayDir, "", syscall.MS_BIND, "")
+	return CreateOverlayByPath(bundlePath, overlayPath)
+}
+
+// CreateOverlayByPath creates a writable overlay based on a directory whose path is specified in the second argument.
+func CreateOverlayByPath(bundlePath string, overlayPath string) error {
+	var err error
+
+	_, err = os.Stat(overlayPath)
 	if err != nil {
-		return fmt.Errorf("failed to bind %s: %s", overlayDir, err)
+		if os.IsNotExist(err) {
+			err := os.Mkdir(overlayPath, 0o755)
+			if err != nil {
+				return fmt.Errorf("failed to create %s: %s", overlayPath, err)
+			}
+		} else {
+			return err
+		}
+	}
+
+	err = syscall.Mount(overlayPath, overlayPath, "", syscall.MS_BIND, "")
+	if err != nil {
+		return fmt.Errorf("failed to bind %s: %s", overlayPath, err)
 	}
 	// best effort to cleanup mount
 	defer func() {
 		if err != nil {
-			syscall.Unmount(overlayDir, syscall.MNT_DETACH)
+			syscall.Unmount(overlayPath, syscall.MNT_DETACH)
 		}
 	}()
 
-	if err = syscall.Mount("", overlayDir, "", syscall.MS_REMOUNT|syscall.MS_BIND, ""); err != nil {
-		return fmt.Errorf("failed to remount %s: %s", overlayDir, err)
+	if err = syscall.Mount("", overlayPath, "", syscall.MS_REMOUNT|syscall.MS_BIND, ""); err != nil {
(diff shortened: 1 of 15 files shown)


pkg/ocibundle/tools/overlay_linux.go
```diff
@@ -84,20 +102,41 @@ func CreateOverlayTmpfs(bundlePath string, sizeMiB int) error {
 }
 
 func prepareOverlay(bundlePath, overlayDir string) error {
+	var err error
+
 	upperDir := filepath.Join(overlayDir, "upper")
-	if err := os.Mkdir(upperDir, 0o755); err != nil {
-		return fmt.Errorf("failed to create %s: %s", upperDir, err)
+	_, err = os.Stat(upperDir)
+	if err != nil {
```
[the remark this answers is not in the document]
preminger: Not sure I understand your intent on this one. Unless I'm missing something, rewriting it with this pattern will have to involve consecutive `if` statements that look like this:

```
if err != nil && !os.IsNotExist(err) {
  return err
}

if err != nil {
	err := os.Mkdir(upperDir, 0o755)
	if err != nil {
		return fmt.Errorf("failed to create %s: %s", upperDir, err)
	}
}
```

(Since, if the error occurred and it _was_ a not-exist scenario, we attempt to create the directory.)

Is that what you had in mind?

NB: I don't think we should invert the order of the two if-blocks here, because it'll create confusion about which `err` value we're looking at at a given point, due to the error checking for `os.Mkdir`. (Technically, `err` gets its own `:=` inside the conditional, so it won't overwrite the outer `err`, but writing code that only works by virtue of that kind of scope subtlety seems bad from a readability perspective, as does starting to mess with `err1` and `err2` variables.)

pkg/ocibundle/tools/overlay_linux.go
```diff
@@ -30,22 +35,47 @@ func CreateOverlay(bundlePath string) error {
...
+		}()
+
+		// Try to perform remount
+		if err = syscall.Mount("", d, "", syscall.MS_REMOUNT|syscall.MS_BIND, ""); err != nil {
+			return fmt.Errorf("failed to remount %s: %s", d, err)
+		}
 	}
 
-	err = prepareOverlay(bundlePath, overlayDir)
+	// Prepare internal structure of overlay dir
```
dtrudg, line 77: This comment is a little misleading, as `prepareOverlay` also performs the mount.

dtrudg (changes requested):
This looks pretty good. Great to have behaviour of the native runtime overlay replicated.

There are some comments on the code for minor things. Some more general points:

(1) With the removal of `oci.RunWrapped` from the oci launcher package we now have some logic for handling overlays that is duplicated and separated between two areas of the code base.  If we aren't using a user namespace, then we invoke `l.RunWithOverlays`. If we are using a user namespace, the work that would be performed by `l.RunWithOverlays` is instead directly performed in `internal/app/singularity/oci_linux.go`.

The first makes overlay handling functionality of the launcher package. The second makes it functionality of the singularity CLI. Let's talk about where it should be in order to make things consistent.

(2) In the e2e tests you've changed a container image that is used widely. This is definitely the type of change that (outside of a test-refactoring PR) should be a separate commit, with justification of the change in the commit message.

Let's jump on a paired session to finish this one up.

pkg/ocibundle/tools/overlay_linux.go
```diff
@@ -79,41 +83,175 @@ func CreateOverlayTmpfs(bundlePath string, sizeMiB int) error {
...
+	}
+
+	return nil
+}
+
+// ensureOverlayDir checks if a directory already exists; if it doesn't, and writable is true, it attempts to create it with the specified permissions.
+func ensureOverlayDir(dir string, createIfMissing bool, createPerm os.FileMode) error {
+	_, err := os.Stat(dir)
+	if dir == "" {
+		panic("ensureOverlayDir on empty dir")
```
wobito, line 211: is a panic the right approach, as the return value is `error` and the callers handle the return error

preminger: Oof, no. This is leftover from some debugging. Thanks for catching it!

pkg/ocibundle/tools/overlay_linux.go
```diff
@@ -79,41 +83,175 @@ func CreateOverlayTmpfs(bundlePath string, sizeMiB int) error {
...
+	if err == nil {
+		return nil
+	}
+
+	if !os.IsNotExist(err) {
+		return err
+	}
+
+	if !createIfMissing {
+		return fmt.Errorf("missing overlay dir %#v", dir)
```
dtrudg, line 223: Nit (which I think I've overlooked before) - it's a little unusual to use %#v for something that's definitely a string. %q is string specific and equivalent here, afaik.

'%q' is used quite widely in the code while '%#v' is not, so prefer following the existing pattern unless there's a specific reason, i.e. quoted / go syntax output of a non-string type is required.

wobito (commented):
> (2) In the e2e tests you've changed a container image that is used widely. This is definitely the type of change that (outside of a test-refactoring PR) should be a separate commit, with justification of the change in the commit message.

To this point, I agree, its unclear to me why the test image is changing.

PR #1622, opened 2023-05-02
oci: Enable --writable-tmpfs behaviour by default

dtrudg:
## Description of the Pull Request (PR):

The `--oci` mode intends to follow behaviour that the native runtime implements when run with `--compat`.

One missing aspect is that `--compat` sets `--writable-tmpfs`, where the container rootfs is made writable with a tmpfs backed overlay.

This PR:

- Introduces a simple wrapping of the `oci run` sub-command as `oci run-wrapped`. This hidden command implements prep / cleanup steps that must take place in a userns for non-root `--oci` execution.
- Switches the oci launcher to callin `oci run-wrapped` instead of `oci-run`.
- Adds a tmpfs based overlay creation function for OCI bundles.
- Includes the tmpfs overlay creation in the `oci run-wrapped` flow.
- Copies the native runtime `--compat` e2e tests to OCI mode.

### This fixes or addresses the following GitHub issues:

 - Fixes #1621 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/main/CONTRIBUTORS.md)

PR #1671, opened 2023-05-11
oci: wrap run in writable tmpfs overlay if there's only read-only overlay(s)

### LINKED ISSUE #1670: bug: oci: when mounting only ro overlay(s), busybox images fail to run
**Version of Singularity**

HEAD of `main`

**Describe the bug**
When using (newly added) `--overlay` in OCI mode, the current `busybox` docker image fails to run, reporting a failed attempt to write to a read-only filesystem, involving `/proc`

**To Reproduce**

```
$ mkdir -p ~/tmp/blank-playground
mkdir: created directory '/home/omer/tmp/blank-playground'

$ cd ~/tmp/blank-playground

$ mkdir ro-overlay-dir
mkdir: created directory 'ro-overlay-dir'

$ singularity shell --oci --fakeroot --overlay ro-overlay-dir:ro docker://busybox
2023/05/11 10:27:02  info unpack layer: sha256:b50100f25006c29bd3a3dd4abacfeb7e9cb61c1a758d07c68fa699a2494fd2df
2023-05-11T14:27:02.286834Z: cannot mkdir `proc`: Read-only file system

$ sudo singularity shell --oci --overlay ro-overlay-dir:ro docker://busybox
2023/05/11 10:27:13  info unpack layer: sha256:b50100f25006c29bd3a3dd4abacfeb7e9cb61c1a758d07c68fa699a2494fd2df
2023-05-11T14:27:13.865141Z: cannot mkdir `proc`: Read-only file system

$
```

**Expected behavior**
Here's the behavior with (as an example) the current `alpine` docker image:

```
$ singularity shell --oci --fakeroot --overlay ro-overlay-dir:ro docker://alpine
2023/05/11 10:27:30  info unpack layer: sha256:08409d4172603f40b56eb6b76240a1e6bd78baa0e96590dc7ff76c5f1a093af2
Singularity>

$ sudo singularity shell --oci --overlay ro-overlay-dir:ro docker://alpine
2023/05/11 10:27:37  info unpack layer: sha256:08409d4172603f40b56eb6b76240a1e6bd78baa0e96590dc7ff76c5f1a093af2
Singularity>

$
```

**OS / Linux Distribution**
Which Linux distribution are you using?
```
NAME="Fedora Linux"
VERSION="37 (KDE Plasma)"
ID=fedora
VERSION_ID=37
VERSION_CODENAME=""
PLATFORM_ID="platform:f37"
PRETTY_NAME="Fedora Linux 37 (KDE Plasma)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:37"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f37/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=37
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=37
SUPPORT_END=2023-11-14
VARIANT="KDE Plasma"
VARIANT_ID=kde
```

**Installation Method**
Built from source (using HEAD of `main`)

preminger:

In OCI mode, if there is one or more overlays specified but they are all read-only, wrap Run in an additional writable tmpfs overlay so that the filesystem in its entirety is writable inside the container (in alignment with the goals of #1622 ).

 - Fixes #1670

### DIFF
--- a/e2e/actions/oci.go
+++ b/e2e/actions/oci.go
@@ -980,7 +980,7 @@ func (c actionTests) actionOciOverlay(t *testing.T) {
 	e2e.EnsureOCIArchive(t, c.env)
 	imageRef := "oci-archive:" + c.env.OCIArchivePath
 
-	for _, profile := range []e2e.Profile{e2e.OCIRootProfile, e2e.OCIFakerootProfile} {
+	for _, profile := range e2e.OCIProfiles {
 		testDir, err := fs.MakeTmpDir(c.env.TestDir, "overlaytestdir", 0o755)
 		if err != nil {
 			t.Fatal(err)
@@ -1040,6 +1040,11 @@ func (c actionTests) actionOciOverlay(t *testing.T) {
 				name:     "NonExistReadonly",
 				args:     []string{"--overlay", filepath.Join(testDir, "my_ro_ol_dir_nonexistent:ro"), imageRef, "echo", "hi"},
 				exitCode: 255,
+			},
+			{
+				name:     "ReadonlyAddsTmpfs",
+				args:     []string{"--overlay", filepath.Join(testDir, "my_ro_ol_dir1:ro"), imageRef, "sh", "-c", "echo this_should_disappear > /my_test_file"},
+				exitCode: 0,
 			},
 			{
 				name:     "SeveralReadonly",
--- a/e2e/docker/docker.go
+++ b/e2e/docker/docker.go
@@ -1027,6 +1027,7 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
 			t.Run("issue 1286", c.issue1286)
 			t.Run("issue 1528", c.issue1528)
 			t.Run("issue 1586", c.issue1586)
+			t.Run("issue 1670", c.issue1670)
 		},
 		// Tests that are especially slow, or run against a local docker
 		// registry, can be run in parallel, with `--disable-cache` used within
--- a/e2e/docker/regressions.go
+++ b/e2e/docker/regressions.go
@@ -288,3 +288,26 @@ func (c ctx) issue1586(t *testing.T) {
 		t.Errorf("TMPDIR is not empty after singularity exited")
 	}
 }
+
+// https://github.com/sylabs/singularity/issues/1670
+// Check that runc/crun can add directories the rootfs before entering the
+// container, by running a container based on busybox that lacks, e.g., /proc
+func (c ctx) issue1670(t *testing.T) {
+	for _, profile := range e2e.OCIProfiles {
+		tmpDir, cleanup := e2e.MakeTempDir(t, c.env.TestDir, fmt.Sprintf("issue1670-%s-", profile.String()), "")
+		t.Cleanup(func() {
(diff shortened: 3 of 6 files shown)


preminger:
This was tantalizingly close to just being a matter of replacing

```
err = f()
```
(at internal/pkg/runtime/launcher/oci/oci_overlay.go:69)

with

```
	if writableOverlayFound {
		err = f()
	} else {
		err = WrapWithWritableTmpFs(f, bundleDir)
	}
```

and calling it a day.

However, in the course of doing that, I found that following #1478 , we were no longer tearing down writable tmpfs overlays correctly. Basically, there is one more mount (the mount of the tmpfs filesystem at the overlay directory) taking place in WrapWithWritableTmpfs vs. what takes place in WrapWithOverlays.

In the pre- #1478 code, mounting of the tmpfs filesystem at the overlay directory _took the place_ of the identity bind-mount we do for all overlays (as part of the bind-mount & remount pattern). So in that case, performing the same sequence of detach-unmounts in both scenarios worked.

But there was a large amount of code duplication there, and now both flows pass through ApplyOverlay() (at pkg/ocibundle/tools/overlay_linux.go:140), allowing the bulk of that code to be reused. However, this means there is an "extra" mount in the tmpfs scenario relative to the regular overlay-dir scenario. That's because ApplyOverlay() doesn't "know" if the overlay-dir it's getting is the root of a filesystem (as it is in the tmpfs scenario) or not (as is probably the case when the overlay-dir is a garden-variety subdirectory), so it goes ahead and does the bind-mount & remount pattern either way.

I'm fairly sure that from a resource perspective, this is unbothersome - after all, if it was, then there would be reason not to go through the bind-mount & remount pattern when it's not strictly necessary, either - but it _does_ require an extra detach-unmount to be performed in the tmpfs case. Handling this correctly meant splitting up what used to be `unmountAndDeleteOverlay(rootFsDir, overlayDir)` into two separate utility functions - `detachMount(rootFsDir)` and `detachAndDelete(overlayDir)` - so that an extra `detachMount(overlayDir)` could be interjected between the two in the case of a tmpfs overlay. This also has the upshot of making our internal utility functions hew more closely to how the underlying system calls work.

Lastly: in the course of figuring this out, I added a bunch of `sylog.Debugf()` statements, which it seems to me should stay part of the code now, since they provide useful info when turned on (and can be turned off like all other `sylog.Debugf()` statements).

dtrudg (changes requested):
Logic looks good to me, but there should be a test that protects us from regressing in the #1670 busybox case (or an example engineered to capture the circumstances of the busybox case).

dtrudg (approved):

PR #1678, opened 2023-05-12
oci: Improve testing of overlay cleanup

preminger:

In working on #1659 and #1671 , it emerged that in some scenarios (esp. when running as root), our cleanup of overlay mounts was incomplete. Importantly, that cleanup process intentionally logs errors and proceeds (rather than erroring out), in order to clean up as much as possible even under error conditions; but that means that we can't rely simply on exit codes to detect problems in the cleanup of overlay mounts.

This PR adds an e2e test specifically dedicated to detecting such problems, by running an OCI-mode container as root, and comparing the number of current mounts before & after the run.

### DIFF
--- a/e2e/actions/oci.go
+++ b/e2e/actions/oci.go
@@ -6,6 +6,7 @@
 package actions
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


e2e/actions/oci.go
```diff
@@ -1074,3 +1076,56 @@ func (c actionTests) actionOciOverlay(t *testing.T) {
...
 	}
 }
+
+// actionOciOverlayTeardown checks that OCI-mode overlays are correctly
+// unmounted even in root mode (i.e., when user namespaces are not involved).
+func (c actionTests) actionOciOverlayTeardown(t *testing.T) {
+	e2e.EnsureOCIArchive(t, c.env)
+	imageRef := "oci-archive:" + c.env.OCIArchivePath
+
+	mountInfoPath := "/proc/self/mountinfo"
```
dtrudg, line 1086: Nit: If there's a string that's definitely a constant, declare it `const` when possible, so that you don't ever accidentally change it.

PR #1687, opened 2023-05-17
Set custom configuration directory via environment variable

Rubo3:
This PR adds a new check for the configuration directory via the `SINGULARITY_CONFIGDIR` environment variable, which is checked before the current user's home directory in `sysfs.configDir()`. This is useful if one does not want `.singularity` in their home directory, and can be used to enforce the XDG config directory of the [XDG Base Directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) defining `SINGULARITY_CONFIGDIR=$XDG_CONFIG_HOME/singularity`.
This environment variable complements `SINGULARITY_CACHEDIR`, which can be used to do the same for the XDG cache directory with `SINGULARITY_CACHEDIR=$XDG_CACHE_HOME/singularity`.

### DIFF
--- a/e2e/cmdenvvars/cmdenvvars.go
+++ b/e2e/cmdenvvars/cmdenvvars.go
@@ -263,10 +263,10 @@ func (c ctx) testSingularityConfigDir(t *testing.T) {
 	// Test plan:
 	//
 	// - create a temporary directory to be the configuration directory
-	// - set the SINGULARITY_CONFIGDIR environment variable
-	//   to that temporary directory
 	// - run 'singularity remote list' to create the remote.yaml
-	//   file inside the configuration directory
+	//   file inside the configuration directory, with the
+	//   SINGULARITY_CONFIGDIR environment variables set to that temporary
+	//   directory
 	// - assert that the file has been created
 	//
 	// If the file is in the temporary directory, it means
@@ -276,13 +276,13 @@ func (c ctx) testSingularityConfigDir(t *testing.T) {
 	configDir, cleanup := setupTemporaryDir(t, c.env.TestDir, "config-dir")
 	defer cleanup(t)
 
-	os.Setenv("SINGULARITY_CONFIGDIR", configDir)
-
+	environ := append(os.Environ(), fmt.Sprintf("SINGULARITY_CONFIGDIR=%s", configDir))
 	c.env.RunSingularity(
 		t,
 		e2e.WithProfile(e2e.UserProfile),
 		e2e.WithCommand("remote"),
 		e2e.WithArgs("list"),
+		e2e.WithEnv(environ),
 		e2e.ExpectExit(0),
 	)
 
--- a/e2e/cmdenvvars/cmdenvvars.go
+++ b/e2e/cmdenvvars/cmdenvvars.go
@@ -270,11 +270,24 @@ func (c ctx) testSingularitySypgpDir(t *testing.T) {
 }
 
 func (c ctx) testSingularityConfigDir(t *testing.T) {
+	// Test plan:
+	//
+	// - create a temporary directory to be the configuration directory
+	// - set the SINGULARITY_CONFIGDIR environment variable
+	//   to that temporary directory
+	// - run 'singularity remote list' to create the remote.yaml
+	//   file inside the configuration directory
+	// - assert that the file has been created
+	//
+	// If the file is in the temporary directory, it means
+	// singularity followed the SINGULARITY_CONFIGDIR environment
+	// variable (checked in pkg/syfs) to set the configuration directory.
+
 	configDir, cleanup := setupTemporaryDir(t, c.env.TestDir, "config-dir")
 	defer cleanup(t)
-	
+
 	os.Setenv("SINGULARITY_CONFIGDIR", configDir)
-	
+
 	c.env.RunSingularity(
 		t,
 		e2e.WithProfile(e2e.UserProfile),
@@ -282,7 +295,7 @@ func (c ctx) testSingularityConfigDir(t *testing.T) {
 		e2e.WithArgs("list"),
 		e2e.ExpectExit(0),
 	)
-	
+
 	remotePath := filepath.Join(configDir, "remote.yaml")
 	if _, err := os.Stat(remotePath); os.IsNotExist(err) {
 		t.Fatalf("failed to find remote.yaml (expected: %s)", remotePath)
@@ -300,6 +313,6 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
 		"SINGULARITY_CACHEDIR":      c.testSingularityCacheDir,
 		"singularity disable cache": c.testSingularityDisableCache,
 		"SINGULARITY_SYPGPDIR":      c.testSingularitySypgpDir,
-		"SINGULARITY_CONFIGDIR":     c.testSingularityConfigDir,
+		"config directory":          c.testSingularityConfigDir,
 	}
 }
--- a/e2e/cmdenvvars/cmdenvvars.go
+++ b/e2e/cmdenvvars/cmdenvvars.go
@@ -6,6 +6,7 @@
 package cmdenvvars
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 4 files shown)


pkg/syfs/syfs.go
```diff
@@ -43,20 +43,26 @@ func ConfigDir() string {
 }
 
 func configDir() string {
-	user, err := user.Current()
-	if err != nil {
-		sylog.Warningf("Could not lookup the current user's information: %s", err)
+	configDir := os.Getenv("SINGULARITY_CONFIGDIR")
+	if configDir == "" {
+		sylog.Warningf("Environment variable SINGULARITY_CONFIGDIR is not set")
```
dtrudg, line 48: This shouldn't be a warning. Most users will not set the environment variable, so we don't want a warning every time singularity is run by them.

pkg/syfs/syfs.go
```diff
@@ -43,20 +43,26 @@ func ConfigDir() string {
...
+
+			return filepath.Join(cwd, singularityDir)
 		}
 
-		return filepath.Join(cwd, singularityDir)
+		return filepath.Join(user.HomeDir, singularityDir)
 	}
-
-	return filepath.Join(user.HomeDir, singularityDir)
+	return configDir
```
dtrudg, line 65: This is a long way from where `configDir` is extracted from the environment variable. It would be better to handle the case that the environment variable is set at the top of the function, returning as soon as possible.

That change would make the flow through the function easier to read, and remove one level of nesting.

pkg/syfs/syfs.go
```diff
@@ -43,20 +43,26 @@ func ConfigDir() string {
 }
 
 func configDir() string {
-	user, err := user.Current()
-	if err != nil {
-		sylog.Warningf("Could not lookup the current user's information: %s", err)
+	configDir := os.Getenv("SINGULARITY_CONFIGDIR")
```
dtrudg, line 46: Ideally, declare a constant for `SINGULARITY_CONFIGDIR` and use it in the Getenv and any error messages.

We have other code that doesn't do this, but going forward we want to avoid hard coding these in multiple locations.

dtrudg (changes requested):
Hi @Rubo3 and thanks for the submission.

I've reviewed the code submitted, and I think a `SINGULARITY_CONFIGDIR` env var is a reasonable thing to introduce.

For the contribution to be accepted there would also need to be a test added to:

https://github.com/sylabs/singularity/blob/main/e2e/cmdenvvars/cmdenvvars.go

... which would set `SINGULARITY_CONFIGDIR`, run some kind of operation that will modify the config dir, and check that the expected files are at the correct location.

Stepping back a bit, the motivation here is to introduce the env var in order to honour the XDG dirs? We do have an aspiration (recorded somewhere... but I can't find it easily... will open an issue), to honour the XDG dirs by default in the next major version release of SingularityCE.

Is that something you would be interested in working on? Ideally we'd use....

* The SINGULARITY_CONFIGDIR / SINGULARITY_CACHEDIR locations if they are set.
* The existing .singularity locations for config and cache if they exist, for backward compatibility.
* The XDG .config / .cache locations otherwise.

Rubo3:
> Is that something you would be interested in working on?

Sure.

pkg/syfs/syfs.go
Rubo3: I thought that, too, but I didn't know if you would prefer a flat flow. Will change that.

dtrudg:
> > Is that something you would be interested in working on?
> 
> Sure.

Great! If you are able to tidy this PR as indicated in the comments then we can merge it for the `SINGULARITY_CONFIGDIR` support. Any work to honour XDG dirs by default should then be in a separate PR... as that'll be a behaviour change.

Rubo3:
> Ideally we'd use....
>
>    The SINGULARITY_CONFIGDIR / SINGULARITY_CACHEDIR locations if they are set.
>    The existing .singularity locations for config and cache if they exist, for backward compatibility.
>    The XDG .config / .cache locations otherwise.

`configDir` also checks the current directory. Do I have to keep it?

pkg/syfs/syfs.go
Rubo3: So it would be `Infof`?

dtrudg: No, please remove it. We don't want any output when singularity is using the default location.

dtrudg:
> > Ideally we'd use....
> > 
> > * The SINGULARITY_CONFIGDIR / SINGULARITY_CACHEDIR locations if they are set.
> > * The existing .singularity locations for config and cache if they exist, for backward compatibility.
> > * The XDG .config / .cache locations otherwise.
> 
> `configDir` also checks for the current directory. Do I have to keep it? It isn't included in the bulleted list above.

It needs to be kept in this PR... here we aren't dealing with any change in behaviour.

For the future PR handling the XDG dirs I think it's worth discussing. I'm not sure many people know about this behaviour, or use it. We'll have to open an issue to properly set out the behaviour needed for the XDG related PR, and we can ask for opinions there.

pkg/syfs/syfs.go
Rubo3: Ok

dtrudg:
> Understood.
> 
> > ... which would set `SINGULARITY_CONFIGDIR`, run some kind of operation that will modify the config dir, and check that the expected files are at the correct location.
> 
> Do you have any suggestions for the operation to perform?

One simple option would be to run `singularity remote list`, as this command will ensure that .singularity/remote.yaml is created.

Rubo3:
I added a test, but it seems I am not able to run them. From the root of the repo, `~/singularity`, with `make -C builddir e2e-test E2E_GROUPS=CMDENVVARS`, I get:

```bash
home.go:122: failed to change working directory: chdir /home/<my-user>/singularity/e2e: no such file or directory
```

Do you know what could be the error?

Rubo3:
@dtrudg the test I addedd passes. What do you think?

e2e/cmdenvvars/cmdenvvars.go
```diff
@@ -259,6 +259,39 @@ func (c ctx) testSingularitySypgpDir(t *testing.T) {
...
+	// - assert that the file has been created
+	//
+	// If the file is in the temporary directory, it means
+	// singularity followed the SINGULARITY_CONFIGDIR environment
+	// variable (checked in pkg/syfs) to set the configuration directory.
+
+	configDir, cleanup := setupTemporaryDir(t, c.env.TestDir, "config-dir")
+	defer cleanup(t)
+
+	os.Setenv("SINGULARITY_CONFIGDIR", configDir)
```
dtrudg, line 279: Because you set the environment variable, but do not unset it, it is causing later test to fail.

The best fix is to add something like...

```
	e2e.WithEnv(append(os.Environ(), fmt.Sprintf("SINGULARITY_CONFIGDIR=%s", configDir))
```

... to the RunSingularity call. This will ensure the environment is only set for that particular singularity invocation.

pkg/syfs/syfs.go
```diff
@@ -43,20 +44,24 @@ func ConfigDir() string {
...
 
+	cwd, err := os.Getwd()
+	if err == nil {
 		return filepath.Join(cwd, singularityDir)
 	}
-
-	return filepath.Join(user.HomeDir, singularityDir)
+	sylog.Warningf("Could not get current working directory: %s", err)
+	
+	return singularityDir
```
dtrudg, line 64: This is getting a bit complex to review, for what should be an isolated change, because it changes the order in which things are considered / returned. 

Because you are just adding an override with an env var, the safest thing to do here is just insert the block that handles the env var at the top of the function, and leave the rest of the function as-is, with no further changes.

I.E. add...

```
	configDir := os.Getenv(configDirEnv)
	if configDir != "" {
		return configDir
	}
```

... to the top, then leave everything else the same as it was.

dtrudg (changes requested):

dtrudg:
The lint check is also failing. Please see https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md for instructions on how to use the `gofumpt` tool to tidy the files that have been changed, so that they pass the lint check.

Rubo3:
Now all the checks pass.

dtrudg (approved):
Thanks for the contribution!

PR #1689, opened 2023-05-17
oci: support --workdir

### LINKED ISSUE #1483: Support `--workdir` in `--oci` mode
```
-W, --workdir string                working directory to be used for /tmp, /var/tmp and $HOME (if -c/--contain was also used)
```

In `--oci` mode, it should be possible to use `--workdir <directory>` to specify that  _directory_ is used as a prefix for a set of directories on the host that will be bound to `/tmp`, `/var/tmp`, and `$HOME` (?) inside  the container. By default in `--oci` mode, these will all be `tmpfs` mounts. With `--workdir` in effect they need to be bind mounts, and we have to create the host directories as the source for those mounts if they don't exist.

Note that `--oci` mode is emulating `--compat`, so the usage comment about `--contain` should be in effect here.

Here is how it works in native mode...

```
02:13 pm $ singularity run --contain --workdir wd docker://alpine 
Singularity> touch $HOME/123
Singularity> touch /tmp/abc
Singularity> touch /var/tmp/987
Singularity> 
dtrudg-sylabs@mini:~
02:14 pm $ tree wd
wd
├── tmp
│   └── abc
└── var_tmp
    └── 987

3 directories, 2 files
```

Note that this probably shows a bug in native mode.... I ran with `--contain`, but `$HOME` wasn't mounted from the `wd/` dir on the host that I specified.

It's worth some investigation to try and ascertain whether `$HOME` should be mounted from `wd/`.... or the bug is simply that the doc string is incorrect / hasn't been updated.

Given we have `--home` separately, I'd guess it's quite likely that the `$HOME` mount isn't intended. However, it might be worth digging down into the history of the issues and PRs from Singularity's past at https://github.com/apptainer/singularity/ to see if there's anything there.

Also, looking through the history of the container_linux.go code around the implementation of workdir might be informative?

https://github.com/sylabs/singularity/blob/65858fd0158fe36e3ef50a7d6c9954c464d91589/internal/pkg/runtime/engine/singularity/container_linux.go#L1836
and
https://github.com/sylabs/singularity/blob/65858fd0158fe36e3ef50a7d6c9954c464d91589/internal/pkg/runtime/engine/singularity/container_linux.go#L1718

preminger:

Provides support for `--workdir` in OCI-mode.

Specifying `--workdir <dir>` means that `/tmp` and `/var/tmp` will be mapped, respectively, to `<workdir>/tmp` and `<workdir>/var_tmp` on the host, rather than to tmpfs storage. If `--scratch <scratchdir>` is used in conjunction with `--workdir`, scratch directories will be mapped to subdirectories nested under `<workdir>/scratch` on the host, rather than to tmpfs storage. (This is line with the behavior of `--scratch` in conjunction with `--workdir` in native mode.)

Also changed the usage string for the `--workdir` option to make it clear that $HOME is unaffected by this option. (The latter can be manipulated separately, using the `--home` option.) This is so both in native mode and OCI mode.

 - Fixes #1483

### DIFF
--- a/e2e/actions/common.go
+++ b/e2e/actions/common.go
@@ -1,0 +1,44 @@
+// Copyright (c) 2023, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package actions
+
+import (
+	"os"
+	"testing"
+
+	"github.com/sylabs/singularity/e2e/internal/e2e"
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
+)
+
+func workspaceDirsGenerator(t *testing.T, hostCanaryDir, hostHomeDir, hostWorkDir, hostCanaryFile string) {
+	e2e.Privileged(func(t *testing.T) {
+		if err := os.RemoveAll(hostCanaryDir); err != nil && !os.IsNotExist(err) {
+			t.Fatalf("failed to delete canary_dir: %s", err)
+		}
+		if err := os.RemoveAll(hostHomeDir); err != nil && !os.IsNotExist(err) {
+			t.Fatalf("failed to delete workspace home: %s", err)
+		}
+		if err := os.RemoveAll(hostWorkDir); err != nil && !os.IsNotExist(err) {
+			t.Fatalf("failed to delete workspace home: %s", err)
+		}
+	})(t)
+
+	if err := fs.Mkdir(hostCanaryDir, 0o777); err != nil {
+		t.Fatalf("failed to create canary_dir: %s", err)
+	}
+	if err := fs.Touch(hostCanaryFile); err != nil {
+		t.Fatalf("failed to create canary_file: %s", err)
+	}
+	if err := os.Chmod(hostCanaryFile, 0o777); err != nil {
+		t.Fatalf("failed to apply permissions on canary_file: %s", err)
+	}
+	if err := fs.Mkdir(hostHomeDir, 0o777); err != nil {
+		t.Fatalf("failed to create workspace home directory: %s", err)
+	}
+	if err := fs.Mkdir(hostWorkDir, 0o777); err != nil {
+		t.Fatalf("failed to create workspace home directory: %s", err)
+	}
+}
--- a/e2e/actions/common.go
+++ b/e2e/actions/common.go
@@ -13,7 +13,7 @@ import (
 	"github.com/sylabs/singularity/internal/pkg/util/fs"
 )
 
-func workspaceDirsGenerator(t *testing.T, hostCanaryDir, hostHomeDir, hostWorkDir, hostCanaryFile string) {
+func mkWorkspaceDirs(t *testing.T, hostCanaryDir, hostHomeDir, hostWorkDir, hostCanaryFile string) {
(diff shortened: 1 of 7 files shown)


e2e/actions/oci.go
```diff
@@ -440,9 +427,23 @@ func (c actionTests) actionOciBinds(t *testing.T) {
 			if !fn(path) {
 				t.Errorf("%s not found on host", path)
 			}
-			if err := os.RemoveAll(path); err != nil {
-				t.Errorf("failed to delete %s: %s", path, err)
-			}
+			// This part needs to be in privileged mode because of the following
```
preminger, line 430: @dtrudg: Please have a look at this comment. This is what was behind the somewhat opaque msgs I sent you on Slack earlier today.

If the behavior described in the comment is what we should be expecting, then it is indeed just a matter of making the e2e test resilient to it, and this should be good to go.

But if this behavior shouldn't be happening (i.e., if the intermediate dirs should be mkdir'd with non-shifted uid/gid, or not mkdir'd at all), then we need to discuss more.

Note also that this is consistent with the behavior of native mode, which also mkdirs missing intermediate directories as needed in this scenario, but there are uid/gid maps in play and they are simply created under the ownership of the user running singularity.

e2e/actions/common.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/sylabs/singularity/e2e/internal/e2e"
+	"github.com/sylabs/singularity/internal/pkg/util/fs"
+)
+
+func workspaceDirsGenerator(t *testing.T, hostCanaryDir, hostHomeDir, hostWorkDir, hostCanaryFile string) {
```
preminger, line 16: The linter started complaining about how many lines of code e2e/actions/actions.go and e2e/actions/oci.go had in common (fair point, linter!), so I refactored this function out and put it in this new file.

It's not "maximalist" refactoring – perhaps more could be refactored out – but as things stand now, there _are_ differences between actions.go and oci.go in some of the other nearly-identical portions of the relevant code. (For example, the difference in `checkHostFn()`, discussed in a separate comment in oci.go).

dtrudg (changes requested):
Looks pretty good to me. One issue that I have found...

`--workdir` accepts a relative `./workdir` path, but then it can't mount `--scratch` dirs:

```
10:08 am $ singularity run --oci --workdir ./workdir --scratch /scratch docker://alpine
2023/05/18 10:08:48  info unpack layer: sha256:8a49fdb3b6a5ff2bd8ec6a86c05b2922a0f7454579ecc07637e94dfd1d0639b6
2023-05-18T09:08:48.911142Z: error stat'ing file `workdir/scratch/scratch`: No such file or directory

10:08 am $ singularity run --oci --workdir $(pwd)/workdir --scratch /scratch docker://alpine
2023/05/18 10:08:57  info unpack layer: sha256:8a49fdb3b6a5ff2bd8ec6a86c05b2922a0f7454579ecc07637e94dfd1d0639b6
~ $ 
```

Native mode actually has a similar issue... so you could say this PR is accurately replicating native mode :-) However, I think the behaviour should probably be fixed here, and filed as a bug for native mode.

```
$ singularity run --contain --workdir ./workdir --scratch /scratch docker://alpine
FATAL:   container creation failed: could not bind scratch directory workdir/scratch/scratch into container: source must be an absolute path
```

dtrudg (approved):

PR #1694, opened 2023-05-18
fix rel. path --workdir with --scratch, add native and oci e2e tests

### LINKED ISSUE #1693: bug: when --workdir is relative path, conjunction with --scratch fails
**Version of Singularity**

Occurs both in 3.11.3 release and in current `main` code.

**Describe the bug**

The `--workdir <dir>` flag supports both absolute and relative paths. There is a special behavior when `--workdir` is used in conjunction with `--scratch`, which is that the scratch dirs are mapped to a directory structure under wherever the specified workdir is. However, when the workdir is given as a relative (rather than absolute path), the scratch dir portion of things doesn't work.

**To Reproduce**

```
$ singularity --version
singularity-ce version 3.11.3
$ singularity run --contain --workdir ./workdir --scratch /scratch docker://alpine
FATAL:   container creation failed: could not bind scratch directory workdir/scratch/scratch into container: source must be an absolute path
```

**Expected behavior**

Here's what happens when the same command line is changed to use an absolute path:

```
$ singularity run --contain --workdir $(pwd)/workdir --scratch /scratch docker://alpine
Singularity>
```

And here's relative workdir paths interacting with scratch dirs in OCI mode (... thanks to @dtrudg 's vigilance) using the current `main` branch version of singularity:

```
$ singularity --version
singularity-ce version 3.11.0+290-g78729e5f1
$ singularity run --oci --workdir ./workdir --scratch /scratch docker://alpine
2023/05/18 11:17:51  info unpack layer: sha256:08409d4172603f40b56eb6b76240a1e6bd78baa0e96590dc7ff76c5f1a093af2
~ $
```

**OS / Linux Distribution**
Which Linux distribution are you using?
```
$ cat /etc/os-release
NAME="Fedora Linux"
VERSION="37 (KDE Plasma)"
ID=fedora
VERSION_ID=37
VERSION_CODENAME=""
PLATFORM_ID="platform:f37"
PRETTY_NAME="Fedora Linux 37 (KDE Plasma)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:37"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f37/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=37
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=37
SUPPORT_END=2023-11-14
VARIANT="KDE Plasma"
VARIANT_ID=kde
```

preminger:

Fixes a bug where `--scratch` failed when used in conjunction with `--workdir` if the latter was given a relative (rather than absolute) path.

Also changes the behavior of `--workdir` code so that an error (rather than a warning) is generated if specified workdir cannot be converted to an absolute path.

Lastly, adds e2e tests for this behavior (`--scratch` in conjunction with relative `--workdir`), in both native and OCI modes.

 - Fixes #1693

### DIFF
--- a/internal/pkg/runtime/engine/singularity/container_linux.go
+++ b/internal/pkg/runtime/engine/singularity/container_linux.go
@@ -1870,7 +1870,7 @@ func (c *container) addTmpMount(system *mount.System) error {
 
 			workdir, err := filepath.Abs(filepath.Clean(workdir))
 			if err != nil {
-				sylog.Warningf("Can't determine absolute path of workdir %s", workdir)
+				return fmt.Errorf("can't determine absolute path of workdir %s: %s", workdir, err)
 			}
 
 			tmpSource = filepath.Join(workdir, tmpSource)
@@ -1946,7 +1946,7 @@ func (c *container) addScratchMount(system *mount.System) error {
 	if hasWorkdir {
 		workdir, err = filepath.Abs(filepath.Clean(workdir))
 		if err != nil {
-			sylog.Warningf("Can't determine absolute path of workdir %s", workdir)
+			return fmt.Errorf("can't determine absolute path of workdir %s: %s", workdir, err)
 		}
 		sourceDir := filepath.Join(workdir, scratchSessionDir)
 		if err := fs.MkdirAll(sourceDir, 0o750); err != nil {
--- a/internal/pkg/runtime/engine/singularity/container_linux.go
+++ b/internal/pkg/runtime/engine/singularity/container_linux.go
@@ -1942,9 +1942,12 @@ func (c *container) addScratchMount(system *mount.System) error {
 
 	workdir := c.engine.EngineConfig.GetWorkdir()
 	hasWorkdir := workdir != ""
-
+	var err error
 	if hasWorkdir {
-		workdir = filepath.Clean(workdir)
+		workdir, err = filepath.Abs(filepath.Clean(workdir))
+		if err != nil {
+			sylog.Warningf("Can't determine absolute path of workdir %s", workdir)
+		}
 		sourceDir := filepath.Join(workdir, scratchSessionDir)
 		if err := fs.MkdirAll(sourceDir, 0o750); err != nil {
 			return fmt.Errorf("could not create scratch working directory %s: %s", sourceDir, err)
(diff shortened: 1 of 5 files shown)


internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -1942,9 +1942,12 @@ func (c *container) addScratchMount(system *mount.System) error {
 
 	workdir := c.engine.EngineConfig.GetWorkdir()
 	hasWorkdir := workdir != ""
-
+	var err error
 	if hasWorkdir {
-		workdir = filepath.Clean(workdir)
+		workdir, err = filepath.Abs(filepath.Clean(workdir))
+		if err != nil {
+			sylog.Warningf("Can't determine absolute path of workdir %s", workdir)
```
dtrudg, line 1949: I believe this needs to be an error return, not a warning. If `filepath.Abs` returns an err, the string it returns into `workdir` will be `""` which will mean we proceed with `workdir` effectively being the cwd. We don't want to have any chance that we create our scratch dir in the cwd, when that's not what the user asked for.

dtrudg (changes requested):

internal/pkg/runtime/engine/singularity/container_linux.go
preminger: Okay - but note that I was just extending what was already present in the native-mode code:

```go
			workdir, err := filepath.Abs(filepath.Clean(workdir))
			if err != nil {
				sylog.Warningf("Can't determine absolute path of workdir %s", workdir)
			}

			tmpSource = filepath.Join(workdir, tmpSource)
			vartmpSource = filepath.Join(workdir, vartmpSource)
```
(internal/pkg/runtime/engine/singularity/container_linux.go:1871)

I'll change it in all four places, then: in {tmp-mounting routine, scratch-mounting routine} x {native code, OCI-mode code}.

dtrudg: :+1: Yep, thanks.

dtrudg (approved):
LGTM. Thanks.

PR #1699, opened 2023-05-23
WIP: oci: support --overlay of bare images

### LINKED ISSUE #1479: Support `--overlay` in `--oci` mode (bare image)
In native mode, a bare image file can be used as the source for `--overlay`. The image file can be:

* In extfs format, in which case the overlay is read-write by default, read-only if requested.
* In squashfs format, in which case the overlay is read-only.

The extfs case is documented at: https://docs.sylabs.io/guides/latest/user-guide/persistent_overlays.html#filesystem-image-overlay

The squashfs case is not adequately documented at present, but sees limited use.

To accomplish image overlay, the image files are first mounted onto a session directory using kernel mounts. After this, an overlay mount is used to overlay onto the container rootfs.

In `--oci` mode we want to avoid privilege escalation from a non-root user, so we cannot perform kernel mounts of squashfs and extfs image files. However, FUSE solutions are available for squashfs and extfs:

* squashfuse - https://github.com/vasi/squashfuse
* fuse2fs - https://man7.org/linux/man-pages/man1/fuse2fs.1.html

In `--oci` mode, implement functionality that when a squashfs/extfs image file is presented to the `--overlay` option:

* Will mount it to a staging directory, using squashfuse or fuse2fs.
* Will overlay the mounted filesystem onto the container rootfs.
* Allows r/w for extfs images, unless `:ro` is supplied as an option in the `--overlay` string.
* Cleanly unmounts the overlay and underlying FUSE filesystems when the container exits.

At this stage we will assume that we are using a kernel new enough to support unprivileged overlays without fuse-overlayfs (which we will need to integrate later).

preminger:

OCI-mode: support mounting of squashfs images (read-only) and extfs images (read-write or read-only) with `--overlay`, alongside the already existing functionality of directory-based overlays.

 - Fixes #1479


pkg/ocibundle/tools/overlay_linux.go
```diff
@@ -8,33 +8,217 @@ package tools
...
+	}
+
+	// Even though fusermount is not needed for this step, we shouldn't
+	// do the squashfuse mount unless we have the necessary tools to
+	// eventually unmount it
+	_, err = bin.FindBin("fusermount")
+	if err != nil {
+		return fmt.Errorf("use of squashfs overlay requires fusermount to be installed: %s", err)
+	}
+	sqshfsDir, err := os.MkdirTemp("", "squashfuse-for-oci-overlay-")
```
dtrudg, line 120: This has the potential to cause a security issue in the future, that we should guard against.

It's similar to, but less serious / guaranteed, than prior CVEs - https://github.com/apptainer/singularity/commit/98ad49a3de9043ea287d65f4d544632ff074e506

`MkdirTemp` is going to create a secure `700` permissions temporary directory. However, when you mount a squashfs directly onto it, the permissions will then be those of the root directory of the squashfs. If these are more open than `700`, other users could access the overlay (if the allow_other FUSE option is employed).

```
10:04 am $ ls -la /tmp | grep mysecuretempdir
drwx------. 1 dtrudg-sylabs dtrudg-sylabs      0 May 24 10:03 mysecuretempdir/

10:04 am $ squashfuse image.sqfs /tmp/mysecuretempdir/

10:05 am $ ls -la /tmp | grep mysecuretempdir
drwxr-xr-x. 20 root          root               0 May 18 09:57 mysecuretempdir/
```

We can either...

* Setup our mount target dirs so that they are all inside a provided directory which is assumed/required to be secure (e.g. the OCI bundle directory). This would require passing more information into the overlay code.
* Create a nested directory in the temporary directory, and perform the mount on it, so the permissions of the tempdir do not change form 700.

At this point the second option is easier. We should move to move toward the first at a later stage, but it can be deferred here.

pkg/ocibundle/tools/overlay_set_linux.go
```diff
@@ -0,0 +1,146 @@
...
+	// WritableOverlay is the directory to be mounted as a writable overlay. The
+	// mount point atop which this will be mounted is left implicit, to be
+	// chosen by whichever function consumes the OverlaySet. Empty value
+	// indicates no writable overlay is to be mounted.
+	WritableOverlay *OverlayItem
+}
+
+// Mount prepares and mounts the entire OverlaySet onto the specified rootfs
+// directory.
+func (ovs OverlaySet) Mount(rootFsDir string) error {
```
dtrudg, line 41: Convention in Go is that the receiver var is a single letter.

dtrudg (commented):
Some initial comments on the WIP.

internal/pkg/util/fs/overlay/overlay_item_linux.go
```diff
@@ -0,0 +1,297 @@
...
+type Item struct {
+	// Type represents what type of overlay this is (from among the values in pkg/image)
+	Type int
+
+	// Writable represents whether this is a writable overlay
+	Writable bool
+
+	// BarePath is the path of the overlay stripped of any colon-prefixed
+	// options (like ":ro")
+	BarePath string
```
dtrudg, line 32: `SourcePath` would probably be clearer when it is subsequently used in code. We're starting to get quite overloaded r.e. the concept of `path of the overlay`, and `Bare` isn't very informative.

In comments here, prefer talking about the `overlay Item`, and not just `overlay`.

internal/pkg/util/fs/overlay/overlay_item_linux.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// DirToMount is the path of the directory that will actually be passed to
+	// the mount system-call when mounting this overlay
+	DirToMount string
```
dtrudg, line 36: The name `DirToMount` and the comment are a bit confusing to me. The value is the directory from which this `Item` should be overlayfs mounted in the `Set` context. However, we may have mounts in the `Item` context, so we need to be very specific when we talk about mounts.

I'd suggest something like `StagingDir`, but mostly want to encourage emphasizing that the mount you are talking about is in the Set context, something more like `... is the directory on which this overlay item is staged, to be used as a source for an overlayfs mount as part of an overlay.Set`

internal/pkg/util/fs/overlay/overlay_item_linux.go
```diff
@@ -0,0 +1,297 @@
...
+		return fmt.Errorf("image %s is of a type that is not currently supported as overlay", overlay.BarePath)
+	}
+
+	return nil
+}
+
+// SetSecureParentDir sets the secure parent-dir in which to create
+// overlay-specific mount directories.
+func (i *Item) SetSecureParentDir(d string) {
+	i.secureParentDir = d
```
dtrudg, line 100: If we're calling it `secureParentDir`, and letting it be set from a public method, I'd anticipate some checks on it when set to ensure it is secure.

Otherwise, if we don't want to enforce secure permissions here, it's just a `parentDir`.

internal/pkg/util/fs/overlay/overlay_item_linux.go
```diff
@@ -0,0 +1,297 @@
...
+	} else if err := analyzeImageFile(&overlay); err != nil {
+		return nil, fmt.Errorf("error encountered while examining image file %s: %s", overlay.BarePath, err)
+	}
+
+	return &overlay, nil
+}
+
+// analyzeImageFile attempts to determine the format of an image file based on
+// its header
+func analyzeImageFile(overlay *Item) error {
```
dtrudg, line 76: This should be modified to be `func (i *Item) analyzeImageFile() error`. Using a method on the struct is nicer than a separate function that modifies a struct passed by pointer, given that the code is entirely concerned with an `Item`.

internal/pkg/util/fs/overlay/overlay_item_linux.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// Best effort to cleanup temporary dir
+	defer func() {
+		if err != nil {
+			sylog.Debugf("Encountered error with current OverlaySet; attempting to remove %q", fuseMountDir)
+			os.Remove(fuseMountDir)
+		}
+	}()
+
+	execCmd := exec.Command(fuseMountCmd, i.BarePath, fuseMountDir)
```
dtrudg, line 211: The FUSE mount of an image will need to be able to mount the image read-only, when the overlay request is read-only.

For squashfs there's no issue... it can only be read only.

For extfs, only one mount can safely be performed in read-write mode. If we want to use a read-only overlay in extfs format across multiple concurrent runs of `singularity`, the image mount must be `ro` as well as the `overlayfs` portion.

dtrudg, line 211: Some debug logging, of the exact command that is called, will probably be very useful in future troubleshooting. FUSE functionality is likely to be 'fun' to prove across all the distributions we must aim to support.

internal/pkg/util/fs/overlay/overlay_item_linux.go
```diff
@@ -0,0 +1,297 @@
...
+// Unmount performs the necessary steps to unmount an individual OverlayItem.
+// Note that this method does not unmount the overlay itself. That happens in
+// OverlaySet.Unmount().
+func (i Item) Unmount() error {
+	switch i.Type {
+	case image.SANDBOX:
+		return i.unmountDir()
+
+	case image.SQUASHFS:
+		fallthrough
```
dtrudg, line 231: Is there a reason you aren't calling unmountFuse here?

internal/pkg/util/fs/overlay/overlay_set_linux.go
```diff
@@ -0,0 +1,143 @@
...
+	ReadonlyOverlays []*Item
+
+	// WritableOverlay is the directory to be mounted as a writable overlay. The
+	// mount point atop which this will be mounted is left implicit, to be
+	// chosen by whichever function consumes the OverlaySet. Empty value
+	// indicates no writable overlay is to be mounted.
+	WritableOverlay *Item
+}
+
+// Mount prepares and mounts the entire OverlaySet onto the specified rootfs
```
dtrudg, line 40:
```suggestion
// Mount prepares and mounts the entire Set onto the specified rootfs
```

(also elsewhere)

dtrudg (commented):
Structure of this is looking good now, with the refactoring of overlay handling out of the OCI specific code :+1:

With regard to the issues noted elsewhere with `fuse2fs` and upper fs feature support, let's make this PR for read-only overlay, ignore the issues for now, and tackle them in a next step.

Because you have decoupled the basic overlay functionality from the OCI bundle <-> launcher code, it should be possible to test it with unit tests relatively easily. Please could you add some basic unit tests over `e2e.Item` and `e2e.Set` functionality, so we aren't entirely dependent on the e2e for this.

preminger:
On this:

> Because you have decoupled the basic overlay functionality from the OCI bundle <-> launcher code, it should be possible to test it with unit tests relatively easily. Please could you add some basic unit tests over `overlay.Item` and `overlay.Set` functionality, so we aren't entirely dependent on the e2e for this.

It's not clear to me at this point what kinds of unit tests can be done on overlay.Item and overlay.Set, that are anything but e2e tests in disguise. The reason I say this is that these objects' code directly (tries to) touch the underlying filesystem (to, e.g., create missing directories, try to determine the type of filesystem an existing image file contains, etc.). As a result, I don't quite see how to do the kind of "dry runs" that unit-testing often involves. (The only exception I currently see to this is the handling of malformed strings, including but not limited to pointing to non-existent paths, in `overlay.Item.NewItemFromString()`.)

Please advise.

dtrudg:
> On this:
> 
> > Because you have decoupled the basic overlay functionality from the OCI bundle <-> launcher code, it should be possible to test it with unit tests relatively easily. Please could you add some basic unit tests over `overlay.Item` and `overlay.Set` functionality, so we aren't entirely dependent on the e2e for this.
> 
> It's not clear to me at this point what kinds of unit tests can be done on overlay.Item and overlay.Set, that are anything but e2e tests in disguise. The reason I say this is that these objects' code directly (tries to) touch the underlying filesystem (to, e.g., create missing directories, try to determine the type of filesystem an existing image file contains, etc.). As a result, I don't quite see how to do the kind of "dry runs" that unit-testing often involves. (The only exception I currently see to this is the handling of malformed strings, including but not limited to pointing to non-existent paths, in `overlay.Item.NewItemFromString()`.)
> 
> Please advise.

Touching the filesystem doesn't necessarily make something an e2e test. The main difference between unit and e2e testing is that:

* A unit test is testing function(s) by invoking them directly - minimizing the amount of code run to observe the correct behaviour of the function.
* An e2e test is invoking functionality as it would be by an end user. I.E. you are testing overlay creation, but through the singularity CLI -> OCI launcher etc.

Aside from this, we try to ensure unit tests don't depend on anything we have to fetch over the network.

It's fair to create (and cleanup) directories and mounts inside a test temporary directory in a unit test.

internal/pkg/util/fs/overlay/overlay_item_linux.go
dtrudg: Yeah sorry about this one... it's obvious today... I must have had brain fade on my first run through :-)

e2e/actions/oci.go
```diff
@@ -23,6 +23,18 @@ import (
...
 )
 
+const (
+	imgTestFilePath    string = "file-for-testing"
+	squashfsTestString string = "squashfs-test-string"
+	extfsTestString    string = "extfs-test-string"
+)
+
+var (
+	imgsPath        = filepath.Join("..", "internal", "pkg", "util", "fs", "overlay", "testdata")
```
dtrudg, line 33: If we're using from the e2e and unit test we should move these file under the top level `test/` directory. I believe there's a place there for fs images already (SIFs used in various tests).

dtrudg (approved):
LGTM. Let's get this merged. If we need to tweak in downstream work on oci-sif & sif overlay partition handling, then we'll do that.

PR #1708, opened 2023-05-26
more robust way of setting IFS to newline

### LINKED ISSUE #1704: bug: Deffile in `singularity inspect` misses trailing "n" unless it appears twice
I found a very curious bug in the output of `singularity inspect -d`:

When the letter "n" is at the end of line in the original definition file it will not show up in the output of the `inspect` command unless it appears at least another time in the line.

I found that when using an `if` in the `%setup` section:
- `if [ ! -e "server/public.gpg" ]; then` --> `if [ ! -e "server/public.gpg" ]; the`
- `if [ ! -e "client/public.gpg" ]; then` --> `if [ ! -e "client/public.gpg" ]; then`

MWE to showcase the issue:
```
BootStrap: docker
From: ubuntu:22.04

%setup
    #an
    #abn
    #nn
    #ann
    #nan
    #anan
```

Build this and inspect it:
```
BootStrap: docker
From: ubuntu:22.04

%setup
    #a
    #ab
    #nn
    #ann
    #nan
    #anan
```

I tested all letters a-z and only "n" is affected.

Tested with 3.7.0 and 3.11.3-jammy

preminger:

As part of the function of `singularity inspect`, an ad-hoc shell script is set up, by concatenating programmatically-generated snippets of shell code. One of these snippets, created in cmd/internal/cli/inspect.go:227, includes a line whose function is to set IFS (the Input Field Separator) _back_ to newline in the scope of the `cat_file()` function, despite being set to ":" (colon) in the global scope (cmd/internal/cli/inspect.go:256).

However, the way this was done until now uses escaping of the newline, as follows: `local IFS=$'\n'`
This escaping apparently doesn't work on certain platforms, resulting in the relevant input being split using the character "n" as the relevant delimiter, rather than using newlines.

I have verified that it fails on Debian 11 "bullseye" on amd64, and the user who reported #1704 encountered the same thing on Ubuntu 22.04 "jammy" on amd64. There is a newer syntax for this - `local IFS=$'...'` - but there is a worry this won't work on older systems; see discussion in [this](https://stackoverflow.com/questions/16831429/when-setting-ifs-to-split-on-newlines-why-is-it-necessary-to-include-a-backspac) StackOverflow post. In the same post, however, it is mentioned that there is a rather simple and possibly more robust way to achieve the same effect, namely, to use a literal newline between double-quotes:

```
      <clipped>
      local IFS="
"
      <clipped>
```

This PR replaces the previous implementation, using `$'\n'`, with the one above.

 - Fixes #1704

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,8 @@
 ### Bug Fixes
 
 - Fix interaction between `--workdir` when given relative path and `--scratch`.
+- Fix dropped "n" characters on some platforms in definition file stored as part
+  of SIF metadata.
 
 ## 3.11.3 \[2023-05-04\]
 
--- a/cmd/internal/cli/inspect.go
+++ b/cmd/internal/cli/inspect.go
@@ -242,7 +242,8 @@ func newCommand(allData bool, appName string, img *image.Image) *command {
 	cat_file() {
 		echo "%[3]s $1:$2"
 
-		local IFS=$'\n'
+		local IFS="
+"
 		while read -r content; do
 			printf "%%s\n" "$content"
 		done < "$2"
--- a/e2e/docker/docker.go
+++ b/e2e/docker/docker.go
@@ -1037,6 +1037,7 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
 		// Regressions
 		"issue 4943": c.issue4943,
 		"issue 5172": c.issue5172,
-		"issue 274":  c.issue274, // https://github.com/sylabs/singularity/issues/274
+		"issue 274":  c.issue274,  // https://github.com/sylabs/singularity/issues/274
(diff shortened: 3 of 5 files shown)


preminger:
fyi - tried cherry-picking _only_ the e2e test commit (`445e9a6155b3ad8377d752280841bd537f5caa10`) onto a clone of `main` to make sure the e2e test indeed fails _without_ the fix itself (`6b02eb5a7347d7c3694806b15f591042027533f6`), and it indeed does, though only on `amd64`. This is consistent with what's reported in #1704 , and the e2e test is based directly on the `.def` file provided by @Flamefire there.

tri-adam (approved):
LGTM

dtrudg (changes requested):
Good catch. Code / test LGTM.

Please could you just add a line to the CHANGELOG highlighting this bugfix. Thanks.

dtrudg (approved):

PR #1730, opened 2023-06-02
oci: fallback to fuse-overlayfs if kernel doesn't support unprivileged overlays

### LINKED ISSUE #1727: Use fuse-overlayfs, where kernel overlay not available (--oci)
At present, the `--oci` mode relies on unprivileged overlay support in the kernel. This is not available on all systems. Introduced in mainline kernel 5.13. it has been backported to some distributions using older kernels, but not others.

We have some code that can check for unprivileged kernel overlay support here:

https://github.com/sylabs/singularity/blob/main/internal/pkg/util/fs/overlay/overlay_linux.go#L141

This check is currently used in native mode. Where unpriv kernel overlay is not supported, native mode will switch to an 'underlay' approach... in which many (many many) bind mounts are used, instead of an overlayfs mount, to assemble the combined view of the rootfs and `--overlay` source.

Underlay is complex and tied to the native runtime session layout. I don't think we want to implement it for `--oci` mode unless we fins a compelling reason to do so.

Instead, the OverlaySet code should use `fuse-overlayfs` (https://github.com/containers/fuse-overlayfs) if kernel unprivileged overlays are not available.

Implement an automatic check, prioritising kernel unprivileged overlay, but falling back to `fuse-overlayfs`.

preminger:

If kernel doesn't support unprivileged overlay mounts, fallback to trying to use `fuse-overlayfs` and `fusermount` for overlay mounting and unmounting.

 - Fixes #1727

### DIFF
--- a/internal/pkg/util/fs/overlay/overlay_item_linux.go
+++ b/internal/pkg/util/fs/overlay/overlay_item_linux.go
@@ -51,7 +51,7 @@ func NewItemFromString(overlayString string) (*Item, error) {
 	splitted := strings.SplitN(overlayString, ":", 2)
 	item.SourcePath, err = filepath.Abs(splitted[0])
 	if err != nil {
-		return nil, fmt.Errorf("error while trying to convert overlay path %q to absolute path: %s", splitted[0], err)
+		return nil, fmt.Errorf("error while trying to convert overlay path %q to absolute path: %w", splitted[0], err)
 	}
 
 	if len(splitted) > 1 {
@@ -71,7 +71,7 @@ func NewItemFromString(overlayString string) (*Item, error) {
 	if s.IsDir() {
 		item.Type = image.SANDBOX
 	} else if err := item.analyzeImageFile(); err != nil {
-		return nil, fmt.Errorf("error encountered while examining image file %s: %s", item.SourcePath, err)
+		return nil, fmt.Errorf("while examining image file %s: %w", item.SourcePath, err)
 	}
 
 	return &item, nil
@@ -160,12 +160,12 @@ func (i *Item) mountDir() error {
 	}
 
 	if err = EnsureOverlayDir(i.StagingDir, false, 0); err != nil {
-		return fmt.Errorf("error accessing directory %s: %s", i.StagingDir, err)
+		return fmt.Errorf("error accessing directory %s: %w", i.StagingDir, err)
 	}
 
 	sylog.Debugf("Performing identity bind-mount of %q", i.StagingDir)
 	if err = syscall.Mount(i.StagingDir, i.StagingDir, "", syscall.MS_BIND, ""); err != nil {
-		return fmt.Errorf("failed to bind %s: %s", i.StagingDir, err)
+		return fmt.Errorf("failed to bind %s: %w", i.StagingDir, err)
 	}
 
 	// Best effort to cleanup mount
@@ -179,7 +179,7 @@ func (i *Item) mountDir() error {
 	// Try to perform remount
 	sylog.Debugf("Performing remount of %q", i.StagingDir)
 	if err = syscall.Mount("", i.StagingDir, "", syscall.MS_REMOUNT|syscall.MS_BIND, ""); err != nil {
-		return fmt.Errorf("failed to remount %s: %s", i.StagingDir, err)
+		return fmt.Errorf("failed to remount %s: %w", i.StagingDir, err)
 	}
 
 	return nil
@@ -192,14 +192,14 @@ func (i *Item) mountWithFuse(fuseMountTool string, additionalArgs ...string) err
 	var err error
 	fuseMountCmd, err := bin.FindBin(fuseMountTool)
 	if err != nil {
-		return fmt.Errorf("use of image %q as overlay requires %s to be installed: %s", i.SourcePath, fuseMountTool, err)
+		return fmt.Errorf("use of image %q as overlay requires %s to be installed: %w", i.SourcePath, fuseMountTool, err)
 	}
 
 	// Even though fusermount is not needed for this step, we shouldn't perform
 	// the mount unless we have the necessary tools to eventually unmount it
 	_, err = bin.FindBin("fusermount")
 	if err != nil {
-		return fmt.Errorf("use of image %q as overlay requires fusermount to be installed: %s", i.SourcePath, err)
+		return fmt.Errorf("use of image %q as overlay requires fusermount to be installed: %w", i.SourcePath, err)
 	}
 
 	// Obtain parent directory in which to create overlay-related mount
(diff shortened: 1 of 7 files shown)


internal/pkg/util/fs/overlay/overlay_item_linux.go
```diff
@@ -47,8 +47,13 @@ type Item struct {
 func NewItemFromString(overlayString string) (*Item, error) {
 	item := Item{Writable: true}
 
+	var err error
 	splitted := strings.SplitN(overlayString, ":", 2)
-	item.SourcePath = splitted[0]
+	item.SourcePath, err = filepath.Abs(splitted[0])
+	if err != nil {
+		return nil, fmt.Errorf("error while trying to convert overlay path %q to absolute path: %s", splitted[0], err)
```
dtrudg, line 54:
```suggestion
		return nil, fmt.Errorf("error while trying to convert overlay path %q to absolute path: %w", splitted[0], err)
```

internal/pkg/util/fs/overlay/overlay_set_linux.go
```diff
@@ -80,9 +101,36 @@ func (s Set) performIndividualMounts() error {
...
-		return fmt.Errorf("failed to mount %s: %s", rootFsDir, err)
+	unprivOls, err := UnprivOverlaysSupported()
+	if err != nil {
+		return fmt.Errorf("error encountered while checking for unprivileged overlay support in kernel: %s", err)
+	}
+
+	if unprivOls {
+		sylog.Debugf("Mounting overlay (via syscall) with rootFsDir %q, options: %q", rootFsDir, options)
+		if err := syscall.Mount("overlay", rootFsDir, "overlay", syscall.MS_NODEV, options); err != nil {
+			return fmt.Errorf("failed to mount %s: %s", rootFsDir, err)
```
dtrudg, line 112:
```suggestion
			return fmt.Errorf("failed to mount %s: %w", rootFsDir, err)
```

internal/pkg/util/fs/overlay/overlay_set_linux.go
```diff
@@ -80,9 +101,36 @@ func (s Set) performIndividualMounts() error {
...
+		}
+	} else {
+		fuseOlFsCmd, err := bin.FindBin("fuse-overlayfs")
+		if err != nil {
+			return fmt.Errorf("kernel does not support unprivileged overlays, and fuse-overlayfs not installed: %s", err)
```
dtrudg, line 117:
```suggestion
			return fmt.Errorf("kernel does not support unprivileged overlays, and fuse-overlayfs not available: %w", err)
```

dtrudg: It may actually be installed, but $PATH is set wrongly, or permissions are incorrect.

internal/pkg/util/fs/overlay/overlay_set_linux.go
```diff
@@ -80,9 +101,36 @@ func (s Set) performIndividualMounts() error {
...
+		if err != nil {
+			return fmt.Errorf("kernel does not support unprivileged overlays, and using fuse-overlayfs fallback requires fusermount to be installed: %s", err)
+		}
+
+		sylog.Debugf("Mounting overlay (via fuse-overlayfs) with rootFsDir %q, options: %q", rootFsDir, options)
+		execCmd := exec.Command(fuseOlFsCmd, "-o", options, rootFsDir)
+		execCmd.Stderr = os.Stderr
+		_, err = execCmd.Output()
+		if err != nil {
+			return fmt.Errorf("failed to mount %s: %s", rootFsDir, err)
```
dtrudg, line 132:
```suggestion
			return fmt.Errorf("failed to mount %s: %w", rootFsDir, err)
```

dtrudg (dismissed):
Looks pretty good to me.

General note - Apologies I have previously been poor at requesting wrapped `%w` error returns (I need to remind myself 😄).

https://go.dev/blog/go1.13-errors

The underlying errors that will, in turn, be returned from a lot of the overlay code are generally going to be system level 'permission denied' / 'operation not supported' things that relate to system configuration, nesting of containers in containers etc. There are circumstances in which we might actually want to inspect the underlying errors to implement workarounds in launcher code.

Because we are in `internal`, a wrapped error doesn't become part of a public API (the caution detailed in the link above).

internal/pkg/util/fs/overlay/overlay_set_linux.go
```diff
@@ -49,7 +61,16 @@ func (s Set) Mount(rootFsDir string) error {
 
 // UnmountOverlay ummounts a Set from a specified rootfs directory.
 func (s Set) Unmount(rootFsDir string) error {
-	if err := DetachMount(rootFsDir); err != nil {
+	unmounterFunc := DetachMount
+	unprivOls, err := UnprivOverlaysSupported()
```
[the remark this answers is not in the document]
dtrudg: Good point. Let's hold that change for this PR.. in case we want to backport onto 3.11, where we should avoid changing those functions if necessary.

internal/pkg/util/fs/overlay/overlay_set_linux.go
```diff
@@ -80,9 +103,36 @@ func (s Set) performIndividualMounts() error {
 func (s Set) performFinalMount(rootFsDir string) error {
 	// Try to perform actual mount
 	options := s.options(rootFsDir)
-	sylog.Debugf("Mounting overlay with rootFsDir %q, options: %q", rootFsDir, options)
-	if err := syscall.Mount("overlay", rootFsDir, "overlay", syscall.MS_NODEV, options); err != nil {
-		return fmt.Errorf("failed to mount %s: %s", rootFsDir, err)
+	unprivOls, err := UnprivOverlaysSupported()
+	if err != nil {
```
dtrudg, line 107:
```suggestion
	if err != ErrNoRootlessOverlay {
```

internal/pkg/util/fs/overlay/overlay_set_linux.go
```diff
@@ -49,7 +61,18 @@ func (s Set) Mount(rootFsDir string) error {
 
 // UnmountOverlay ummounts a Set from a specified rootfs directory.
 func (s Set) Unmount(rootFsDir string) error {
-	if err := DetachMount(rootFsDir); err != nil {
+	unprivOls, err := UnprivOverlaysSupported()
+	if err != nil {
```
dtrudg, line 65:
```suggestion
	if err != ErrNoRootlessOverlay {
```

dtrudg (changes requested):
This is actually failing to follow the required path on a system without unprivileged overlay support (RHEL7 in this case)...

```
# Implied writable-tmpfs overlay
$ singularity run --oci docker://alpine
2023/06/06 09:43:46  info unpack layer: sha256:8a49fdb3b6a5ff2bd8ec6a86c05b2922a0f7454579ecc07637e94dfd1d0639b6
FATAL:   while checking for unprivileged overlay support in kernel: rootless overlay not supported by kernel

# Explicit directory overlay
$ singularity run --oci --overlay /tmp/overlay docker://alpine
2023/06/06 09:44:17  info unpack layer: sha256:8a49fdb3b6a5ff2bd8ec6a86c05b2922a0f7454579ecc07637e94dfd1d0639b6
FATAL:   while checking for unprivileged overlay support in kernel: rootless overlay not supported by kernel
```

Either the error checks where `UnprivOverlaysSupported` need to be changed, or you could modify the function so that in the case of `ErrNoRootlessOverlay` it in turn returns `false, nil`.

Other than that, looks good. With the changes applied it's working on RHEL7.

internal/pkg/util/fs/overlay/overlay_set_linux.go
preminger: Thanks for the catch! (And the work that no doubt went into setting up a RHEL7 environment in which to do so.) But with the suggested change, if the error _is_ nil (as in systems with unprivileged overlay support), it'll go into the error block, right? I think it would be better to change UnprivOverlaysSupported() so that it returned nil in the error field under the relevant circumstances.

internal/pkg/util/fs/overlay/overlay_set_linux.go
preminger: (As above.)

PR #1740, opened 2023-06-06
oci: support for writable extfs img overlay via fuse-overlayfs

### LINKED ISSUE #1728: Support writable bare image overlays (--oci)
Depends on #1727 

In #1699 read-only support for `--overlay` of bare images was added to `--oci` mode via `squashfuse` / `fuse2fs`.

Read-write extfs overlays are currently blocked as kernel unprivileged overlayfs mounts do not support a FUSE upperdir (at least as provided by `fuse2fs`).

#1727 intends to add support for using `fuse-overlayfs` where kernel overlayfs mounts are not supported. In the case that we are mounting a writable extfs overlay, we could also use `fuse-overlayfs` since it does support a FUSE upperdir.

preminger:

Adds support for mounting extfs images as read-write overlays, by using `fuse2fs` (if available) to mount the image followed by `fuse-overlayfs` (if available) to mount it as a read-write overlay.

 - Fixes #1728


e2e/actions/oci.go
```diff
@@ -1236,8 +1243,66 @@ func (c actionTests) actionOciOverlay(t *testing.T) {
...
+				exitCode: 255,
+			},
+			{
+				name:         "WritableExtfs",
+				args:         []string{"--overlay", writableExtfsImgPath, imageRef, "sh", "-c", "echo my_test_string > /my_test_file"},
+				requiredCmds: []string{"fuse2fs", "fuse-overlayfs", "fusermount"},
+				exitCode:     0,
+			},
+			{
+				name:         "WritableExtfsRevisit",
```
dtrudg, line 1255: We need to revisit the `rw` overlay in `ro` mode too, so we catch issue (2) in my general comments.

dtrudg (changes requested):
I'm afraid that this isn't operating in the same way as the native mode, in two significant ways. For the examples below, the overlay was created with:

```
singularity overlay create --size 1024 /tmp/test.img
```

1. Permissions in effect are such that in `--oci` mode I cannot create a file in the root directory `/`, which I should be able to do.

```
$ singularity run --oci --overlay /tmp/test.img:rw docker://alpine
2023/06/07 09:24:16  info unpack layer: sha256:8a49fdb3b6a5ff2bd8ec6a86c05b2922a0f7454579ecc07637e94dfd1d0639b6
unknown argument ignored: lazytime
~ $ touch /bob
touch: /bob: Permission denied

$ singularity run --overlay /tmp/test.img:rw docker://alpine
Singularity> touch /bob
Singularity> 
```

Note that in `--oci` mode this does work for directories, so it's not a generic issue for all rw overlays.. so is definitely within the scope of this PR.

```
$ singularity run --oci --overlay overlaydir:rw docker://alpine
2023/06/07 09:28:26  info unpack layer: sha256:8a49fdb3b6a5ff2bd8ec6a86c05b2922a0f7454579ecc07637e94dfd1d0639b6
~ $ touch /bob
~ $ 
```

2. If I use an overlay `rw` and then later use it `ro` I expect to see the created contents overlayed. In `--oci` mode, the `upper` / `work` are visible:

```
# In native mode... note the `/bob` file present.
 $ singularity exec --overlay /tmp/test.img:ro docker://alpine ls /
bin          environment  lib          opt          run          srv          usr
bob          etc          media        proc         sbin         sys          var
dev          home         mnt          root         singularity  tmp

# In --oci mode, note the upper / work directories, and `/bob` is inside `/upper`
$ singularity exec --oci --overlay /tmp/test.img:ro docker://alpine ls /
2023/06/07 09:33:10  info unpack layer: sha256:8a49fdb3b6a5ff2bd8ec6a86c05b2922a0f7454579ecc07637e94dfd1d0639b6
bin         lib         opt         sbin        upper
dev         lost+found  proc        srv         usr
etc         media       root        sys         var
home        mnt         run         tmp         work
```

This is actually a generic issue that applies to `--oci` overlays in general. Apologies I didn't catch it with the directory overlay. It is up to you whether to address it here, or as a separate issue / PR.

internal/pkg/util/fs/overlay/overlay_item_linux.go
```diff
@@ -155,6 +155,18 @@ func (i *Item) Mount() error {
 	return nil
 }
 
+// GetMountDir returns the path to the directory that will actually be mounted
+// for this overlay. For squashfs overlays, this is equivalent to the
+// Item.StagingDir field. But for all other overlays, it is the "upper"
+// subdirectory of Item.StagingDir.
+func (i Item) GetMountDir() string {
```
dtrudg, line 162: This still isn't quite right. If I am overlaying a directory read-only then it doesn't have to have an upper. If it doesn't have an upper, then the directory content should be overlayed directly:

```
$ ls overlaydirro
bob

$ singularity exec -u --overlay overlaydirro:ro docker://alpine ls /
bin          environment  lib          opt          run          srv          usr
bob          etc          media        proc         sbin         sys          var
dev          home         mnt          root         singularity  tmp
INFO:    Cleaning up image...

$ singularity exec --oci --overlay overlaydirro:ro docker://alpine ls /
2023/06/08 08:59:06  info unpack layer: sha256:8a49fdb3b6a5ff2bd8ec6a86c05b2922a0f7454579ecc07637e94dfd1d0639b6
FATAL:   failed to mount /tmp/oci-bundle3793341770/rootfs: no such file or directory
```

The logic missing here from the native engine is...

https://github.com/sylabs/singularity/blob/60a6afb2184a8684ea1c11cfcf73650d85294e74/internal/pkg/runtime/engine/singularity/container_linux.go#L1050

e2e/actions/oci.go
```diff
@@ -1316,21 +1423,79 @@ func countLines(path string) (int, error) {
...
 	}
-
-	return true
 }
 
 // Make sure --workdir and --scratch work together nicely even when workdir is a
 // relative path. Test needs to be run in non-parallel mode, because it changes
 // the current working directory of the host.
 func (c actionTests) ociRelWorkdirScratch(t *testing.T) {
-	e2e.EnsureOCIArchive(t, c.env)
```
dtrudg, line 1333: Not sure why this has been removed?

e2e/docker/regressions.go
```diff
@@ -344,6 +345,9 @@ func (c ctx) issue1670(t *testing.T) {
 			}
 		})
 
+		dirs.MkdirOrFatal(t, filepath.Join(tmpDir, "upper"), 0o777)
```
dtrudg, line 348: Generally, the need to add additional operations to get an existing test to pass is a good pointer that you need to check, empirically, the behavior of OCI mode vs native mode.

It might actually be necessary, but in those cases it'd also be good to leave a concise justification for reviewers.

internal/pkg/util/fs/overlay/overlay_linux.go
```diff
@@ -31,7 +31,7 @@ const (
 	upperDir
 )
 
-type fs struct {
+type filesys struct {
```
dtrudg, line 34: Please don't change unless necessary, as I don't think it's used here.

If it is necessary to change it, keep it consistent. `incompatibleFs` should now be `incompatibleFilesys`.

internal/pkg/util/fs/overlay/overlay_set_linux.go
```diff
@@ -108,22 +110,26 @@ func (s Set) performFinalMount(rootFsDir string) error {
...
 	}
 
-	if unprivOls {
+	useKernelMount := unprivOls && !s.hasWritableExtfsImg()
+
+	if useKernelMount {
+		flags := uintptr(syscall.MS_NODEV)
 		sylog.Debugf("Mounting overlay (via syscall) with rootFsDir %q, options: %q", rootFsDir, options)
-		if err := syscall.Mount("overlay", rootFsDir, "overlay", syscall.MS_NODEV, options); err != nil {
+		sylog.Debugf("overlay mount flags: %#v", flags)
```
dtrudg, line 118: Nit: add to the related Debugf call above, rather than adding a new message.

dtrudg (changes requested):

internal/pkg/util/fs/overlay/overlay_linux.go
preminger: Several tests in internal/pkg/util/fs/overlay/overlay_item_linux_test.go and internal/pkg/util/fs/overlay/overlay_set_linux_test.go now use fs.CopyFile() from internal/pkg/util/fs/helper.go. While in principle one could have imported "github.com/sylabs/singularity/internal/pkg/util/fs" as <some other name>, in practice the fs.CopyFile() usage is widespread in our code, and I think somebody seeing something like filesys.CopyFile() in only this place would get the impression that this was something other than the ordinary fs.CopyFile() usage.

I'll go ahead and change `incompatibleFilesys` accordingly.

e2e/actions/oci.go
preminger: Me neither... 🤔

PR #1747, opened 2023-06-08
separate out .Raw and .FullRaw in types.Definition

preminger:

When constructing a `types.Definition` object (pkg/build/types/definition.go:18), the function ParseDefinitionFile() (pkg/build/types/parser/deffile.go:419) populates the object's `.Raw` field with the raw bytes _for that build stage_. In a single-stage definition file, this amounts to the entire definition file. But when dealing with multistage builds, this will only contain the raw bytes _for that stage_.

Instead of manually overwriting the `.Raw` field of the last stage in the `[]types.Definition` slice in parser.All() (pkg/build/types/parser/deffile.go:449) so that it contains the raw bytes of the entire definition, as was the behavior until now, this PR introduces a new field `FullRaw` in `types.Definition` that always points to the raw bytes for the entire definition file.

When looking at a `types.Definition` object, one can now be sure that the `.Raw` field contains only the bytes for that build stage (which wasn't true for `types.Definition` representing the last stage in a multistage build, previously), and that the `.FullRaw` field contains the raw bytes for the entire definition file (which was inaccessible from the first (n-1) stages of an n-stage build, previously).


pkg/build/types/parser/deffile_test.go
```diff
@@ -216,10 +217,7 @@ func TestParseDefinitionFile(t *testing.T) {
 			}
 
 			if !reflect.DeepEqual(defTest, defCorrect) {
-				b, _ := json.MarshalIndent(defCorrect, "", "  ")
-				t.Logf("Expected:\n%s", string(b))
-				b, _ = json.MarshalIndent(defTest, "", "  ")
-				t.Logf("Got:\n%s", string(b))
+				assert.DeepEqual(t, defTest, defCorrect)
```
dtrudg, line 220: Not clear why this doesn't replace the reflect.DeepEqual entirely?

pkg/build/types/parser/deffile_test.go
```diff
@@ -441,6 +439,9 @@ func TestParseAll(t *testing.T) {
 			}
 
 			if !reflect.DeepEqual(defTest, defCorrect) {
+				t.Logf("tt.jsonPath: %s", tt.jsonPath)
+				json.NewEncoder(os.Stderr).Encode(defTest)
+				assert.DeepEqual(t, defTest, defCorrect)
```
dtrudg, line 444: Not clear why this doesn't replace the reflect.DeepEqual entirely?

pkg/build/types/definition.go
```diff
@@ -20,7 +20,15 @@ type Definition struct {
...
 	CustomData map[string]string `json:"customData"`
-	Raw        []byte            `json:"raw"`
+
+	// StageRaw contains the raw data for a particular stage of a multistage
+	// definition file, which amounts to the entire definition file if there is
+	// only one build stage.
+	StageRaw []byte `json:"stageraw"`
+
+	// FullRaw contains the raw data for the entire definition.
+	FullRaw []byte `json:"raw"`
```
dtrudg, line 30: If the content of `FullRaw` is now different to `Raw` before, then I'd prefer the serialized name changes to `fullRaw` too, as an indicator they aren't the same.

Do we output the full serialization of `Definition` anywhere, or is it just used in tests?

pkg/build/types/parser/deffile.go
```diff
@@ -417,12 +417,15 @@ func doHeader(h string, d *types.Definition) error {
 // and parse it into a Definition struct or return error if
 // the definition file has a bad section.
 func ParseDefinitionFile(r io.Reader) (d types.Definition, err error) {
-	d.Raw, err = io.ReadAll(r)
+	d.FullRaw, err = io.ReadAll(r)
+	defer func() {
+		d.StageRaw = d.FullRaw
```
dtrudg, line 422: This needs an explanatory comment, that hints at where in the parsing flow this fires, and why it is needed.

dtrudg (commented):
I think this is the right path to take to tidy things up. Couple of comments / questions left.

pkg/build/types/definition.go
preminger: This was done in the interest of backward compatibility - since, for single-stage def files, the content of `FullRaw` _is_ the same as what the content of `Raw` was before, and there is still code in the remote-build area that treats all def files as single-stage even if they're not. By which I mean: code that calls parser.ParseDefinitionFile() instead of calling parser.All() and letting the latter call the former as necessary.

An example is: cmd/internal/cli/build.go:445

```go
	if isValid {
		sylog.Debugf("Found valid definition: %s\n", spec)
		// File exists and contains valid definition
		var defFile *os.File
		defFile, err = os.Open(spec)
		if err != nil {
			return types.Definition{}, err
		}

		defer defFile.Close()

		return parser.ParseDefinitionFile(defFile)
	}
``` 

So I was aiming for as much backward-compatibility as we could muster. This is not a concern with the names of the struct fields themselves, because the compiler and linter will tell you if there's still a piece of code floating around trying to access the now-nonexistent `.Raw` field. But there's no such automatic checking of extant JSON files that might be getting used somewhere. (This may be an unwarranted concern, but I don't yet have a grasp of the totality of the system sufficient to assure myself that it's definitely _not_ the case.)

pkg/build/types/parser/deffile.go
preminger: Oh, good catch - this is actually an artifact of a problem that was rendered moot by other changes. I'll replace it with some much more straightforward code.

pkg/build/types/definition.go
dtrudg: I see where you are going here, but:

- If I'm not completely misunderstanding things, it is *not* the same now for multi-stage.... so this is a breaking change in the meaning of a field in a public JSON structure, in a subset of its usage.
- This is targeting a major release, so breaking changes are okay but...
- Changing the field name makes it much more obvious to anyone consuming the JSON that there has been a breaking change they need to consider.

On that basis I'd still prefer to change the name in the JSON. @EmmEff might agree or disagree, so let's hold on it until he has given input.

dtrudg: Discussed at length and addressed out of band :-)

PR #1751, opened 2023-06-12
fix: add missing cast in memoryusage calculation

preminger:

Adds explicit cast to `uint64` of `Totalram` field in SysInfo system call. This solves a compilation problem on certain versions of Raspberry PI.

This is a pick of https://github.com/apptainer/apptainer/pull/1347

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@
 - Fix dropped "n" characters on some platforms in definition file stored as part
   of SIF metadata.
 - Pass STDIN to `--oci` containers correctly, to fix piping input to a container.
+- Fix memory usage calculation during singularity compilation on RaspberryPi.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -63,6 +63,7 @@
 - Fix dropped "n" characters on some platforms in definition file stored as part
   of SIF metadata.
 - Pass STDIN to `--oci` containers correctly, to fix piping input to a container.
+- Fix memory usage calculation during singularity compilation on RaspberryPi.
```
dtrudg, line 66:
```suggestion
- Fix compilation on 32-bit systems.
```

Let's change the original message here. It's not really important exectly where in the code the issue is. Also, it's for 32-bit systems in general, rather than just Raspberry Pi.

dtrudg (approved):

Issue #696, opened 2022-04-07
addHomeMount causes fatal container creation error, with non-setuid and winbindd


SamStudio8:
**Version of Singularity**
```
$ singularity --version
singularity-ce version 3.9.2
```

**Describe the bug**
When running a demo container I get the error: `container creation failed: failed to add  as session directory: path . is not an absolute path`. I have added several `sylog.Infof` statements to debug and have discovered this is the end result of `lookupUnixId` failing to get my user information to populate the User object with buildUser. The home source and dest are left unpopulated (presumably as empty strings or nil?) and this ultimately leads to a failure to mount the path and causes the container creation to abort (see below).

```
INFO    [U=27900,P=36117]  create()                      About to addHomeMount
INFO    [U=27900,P=36117]  addHomeMount()                addHomeMount
INFO    [U=27900,P=36117]  getHomePaths()                SAM! About to call user.CurrentOriginal
INFO    [U=27900,P=36117]  HostUID()                     SAM! HostUID
INFO    [U=27900,P=36117]  HostUID()                     SAM! HostUID set currentUID to 27900 from os.Getuid
INFO    [U=27900,P=36117]  HostUID()                     SAM! HostUID reading data: [27900 27900 1]
INFO    [U=27900,P=36117]  HostUID()                     SAM! HostUID returning uid=27900, containerID=27900
INFO    [U=27900,P=36117]  CurrentOriginal()             SAM! Calling CurrentOriginal with uid=27900
INFO    [U=27900,P=36117]  GetPwUID()                    SAM! Calling GetPwUID on uid=27900
INFO    [U=27900,P=36117]  lookupUnixUid()               SAM! In lookupUnixId with uid=27900
INFO    [U=27900,P=36117]  retryWithBuffer()             SAM! Trying func
INFO    [U=27900,P=36117]  lookupUnixUid()               SAM! In lookupUnixId with err=no such file or directory
DEBUG   [U=27900,P=36117]  getHomePaths()                SAM! Error in getHomePaths: user: lookup userid 27900: no such file or directory
INFO    [U=27900,P=36117]  addHomeStagingDir()           SAM! Attempting to addHomeStagingDir with source= dest=
VERBOSE [U=27900,P=36117]  CleanupContainer()            Removing image tempDir /tmp/rootfs-2414195454
INFO    [U=27900,P=36117]  CleanupContainer()            Cleaning up image...
[...]

PR #1368, opened 2023-02-24
e2e: refactor test image handling 

dtrudg:
## Description of the Pull Request (PR):

Fetch the OCI archive image used in tests from Docker Hub, so that it will match the host architecture.

### This fixes or addresses the following GitHub issues:

 - Fixes #1364

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/main/CONTRIBUTORS.md)

PR #1756, opened 2023-06-12
Lookup and store user/group information in stage one prior to creating namespaces (apptainer #1368)

preminger:

Lookup and store user/group information in stage one prior to entering any namespaces to fix issue with winbind not correctly lookup user/group information when using user namespace.

This is a pick of https://github.com/apptainer/apptainer/pull/1368

Fixes: #696

### DIFF
--- a/internal/pkg/runtime/engine/singularity/container_linux.go
+++ b/internal/pkg/runtime/engine/singularity/container_linux.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	osuser "os/user"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -1456,13 +1457,14 @@ func (c *container) getHomePaths() (source string, dest string, err error) {
 		dest = filepath.Clean(c.engine.EngineConfig.GetHomeDest())
 		source, err = filepath.Abs(filepath.Clean(c.engine.EngineConfig.GetHomeSource()))
 	} else {
-		pw, err := user.CurrentOriginal()
-		if err == nil {
-			source = pw.Dir
-			if c.engine.EngineConfig.GetFakeroot() {
+		source = c.engine.EngineConfig.JSON.UserInfo.Home
+		if source != "" {
+			if c.engine.EngineConfig.GetFakeroot() || os.Getuid() == 0 {
+				// Mount user home directory onto /root for
+				//  any root-mapped namespace
 				dest = "/root"
 			} else {
-				dest = pw.Dir
+				dest = source
 			}
 		}
 	}
@@ -2001,7 +2003,7 @@ func (c *container) addIdentityMount(system *mount.System) error {
 		if err != nil {
 			sylog.Warningf("%s", err)
 		} else {
-			content, err := files.Passwd(passwd, home, uid)
+			content, err := files.Passwd(passwd, home, uid, c)
 			if err != nil {
 				sylog.Warningf("%s", err)
 			} else {
@@ -2024,7 +2026,7 @@ func (c *container) addIdentityMount(system *mount.System) error {
 
 	if c.engine.EngineConfig.File.ConfigGroup {
 		group := filepath.Join(rootfs, "/etc/group")
-		content, err := files.Group(group, uid, c.engine.EngineConfig.GetTargetGID())
+		content, err := files.Group(group, uid, c.engine.EngineConfig.GetTargetGID(), c)
 		if err != nil {
 			sylog.Warningf("%s", err)
 		} else {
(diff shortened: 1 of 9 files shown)


internal/pkg/util/fs/files/passwd.go
```diff
@@ -65,6 +65,7 @@ func Passwd(path string, home string, uid int, customLookup UserGroupLookup) (co
 
 	sylog.Verbosef("Creating template passwd file and injecting user data: %s", path)
 	userExists := false
+	uid = int(pwInfo.UID)
```
dtrudg, line 68: I'm a little confused here...

What's the circumstance in which `pwInfo.UID` now does not equal the `uid` we were passed in?

dtrudg (changes requested):
I think we need to talk this one through a bit, as I think the adaptation to our handling of the `/etc/passwd` line has introduced a behaviour change when singularity is running in a root-mapped user namespace provided externally. (Could also be from the cherry-picked code though.. I didn't bisect yet).

This might not be a super common situation, but it is definitely something where we should avoid a behaviour change if we can.

internal/pkg/util/fs/files/passwd.go
preminger: I found exactly one such circumstance (so far): it arose during the ContainerAsFakeroot e2e test (in the ACTIONS group).

internal/pkg/runtime/engine/singularity/container_linux.go
```diff
@@ -1638,9 +1638,9 @@ func (c *container) getHomePaths() (source string, dest string, err error) {
 	} else {
 		source = c.engine.EngineConfig.JSON.UserInfo.Home
 		if source != "" {
-			if c.engine.EngineConfig.GetFakeroot() || os.Getuid() == 0 {
```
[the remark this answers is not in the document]
dtrudg: Okay, thanks. I hadn't bisected and was just looking at that commit. I guess it's a regression in Apptainer too... but we'll have to confirm.

internal/pkg/util/fs/files/passwd.go
dtrudg: Hmmm. It's odd that it would only come out in a FUSE related test... and a single one at that. It feels like something is wrong with the cherry-picked change, or alternatively that how it's working in Apptainer isn't really meshing with changes we have.

Looking through I can't see anything obvious. I'm going to need to spend a while on this to understand exactly what's going on.... hopefully tomorrow 🤞

preminger: Possibly worth noting is that the test in question makes use of `e2e/testdata/sshfs.def`, and we've seen before (in the very PR having to do with Passwd(),namely #1531 ) an edge-case that was surfaced by precisely this definition file. The edge-case in question was, at the time, in the situation where the container lacked `/bin/bash`, but the run environment did have `/bin/bash` installed, the shell field of the root user in the container passwd file would get overwritten with `/bin/bash` creating a bricked container. This is obviously not _that_ problem, but it might be worthwhile to think what else might be special about this container...

internal/pkg/runtime/engine/singularity/container_linux.go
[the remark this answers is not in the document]
dtrudg: Confirmed it's a regression in Apptainer too, and nothing to do with this change, as you point out.

See https://github.com/apptainer/apptainer/issues/1450 there. 

I kinda see whats happening but didn't look at a fix yet. I'll come back to it next week.

internal/pkg/util/fs/files/passwd.go
dtrudg: Having looked at it properly now, I think it's better to address this further up in the code. The root cause is that we are re-writing the `root` line for custom `--home` after your revamp of the passwd handling. Therefore it's the root passwd info we actually need, so we should pull that out by target[UG]ID = 0 if `--fakeroot` mode is active.

Apptainer doesn't re-write the `root` line, so it doesn't need to do this.

This should work...

```diff
diff --git a/internal/pkg/runtime/engine/singularity/prepare_linux.go b/internal/pkg/runtime/engine/singularity/prepare_linux.go
index 47437dae2..e4e92e2b5 100644
--- a/internal/pkg/runtime/engine/singularity/prepare_linux.go
+++ b/internal/pkg/runtime/engine/singularity/prepare_linux.go
@@ -130,7 +130,7 @@ func (e *EngineOperations) PrepareConfig(starterConfig *starter.Config) error {
 
        uid := e.EngineConfig.GetTargetUID()
        gids := e.EngineConfig.GetTargetGID()
-       useTargetIDs := false
+       useTargetIDs := e.EngineConfig.GetFakeroot()
 
        if os.Getuid() == 0 && (uid != 0 || len(gids) > 0) {
                starterConfig.SetTargetUID(uid)
diff --git a/internal/pkg/util/fs/files/passwd.go b/internal/pkg/util/fs/files/passwd.go
index 00050d065..0a23c2b97 100644
--- a/internal/pkg/util/fs/files/passwd.go
+++ b/internal/pkg/util/fs/files/passwd.go
@@ -65,7 +65,6 @@ func Passwd(path string, home string, uid int, customLookup UserGroupLookup) (co
 
        sylog.Verbosef("Creating template passwd file and injecting user data: %s", path)
        userExists := false
-       uid = int(pwInfo.UID)
        for i, line := range lines {
                if line == "" {
                        continue
```

Minimally verified against the `fuse_mount` tests only...

```
--- PASS: TestE2E/PAR/ACTIONS/fuse_mount (16.19s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/HostDaemonAsRoot (0.17s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/HostAsRoot (0.16s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/ContainerDaemonAsRoot (0.15s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/ContainerAsRoot (0.16s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/HostDaemonAsUser (0.18s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/HostAsUser (0.15s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/ContainerDaemonAsUser (0.15s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/ContainerAsUser (0.15s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/HostDaemonAsUserNamespace (0.22s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/HostAsUserNamespace (0.16s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/ContainerDaemonAsUserNamespace (0.16s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/ContainerAsUserNamespace (0.15s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/HostDaemonAsFakeroot (0.16s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/HostAsFakeroot (0.15s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/ContainerDaemonAsFakeroot (0.17s)
                --- PASS: TestE2E/PAR/ACTIONS/fuse_mount/ContainerAsFakeroot (0.15s)
```

internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -1440,3 +1446,51 @@ func (e *EngineOperations) loadImage(path string, writable bool) (*image.Image,
 
 	return imgObject, imgErr
 }
+
+func (e *EngineOperations) setUserInfo(useTargetIDs bool) {
+	var gids []int
+
+	pw, err := user.Current()
```
dtrudg, line 1453: Having looked at things... I think that using `user.CurrentOriginal` here, instead, will revert the behaviour to previous w.r.t. the unshare observation. That's what was used previously at `getHomePaths` in container_linux.go

dtrudg (approved):
This LGTM.

I'd appreciate it if you could `rebase -i` so that the pick from apptainer is one commit, and everything else is squashed to a second commit... before merging.

PR #1757, opened 2023-06-12
--no-setgroups option for --fakeroot

### LINKED ISSUE #1748: Allow disabling of setgroups with `--fakeroot`

When a user's `$HOME` is at a location restricted by permissions that depend on supplementary group membership it is not possible to use `--fakeroot` easily. The `$HOME` mount into the initial user/mount namespace will fail, because we use `setgroups` to drop any group other than the (fake) `0`. 

Because of the way `--fakeroot` works, `$HOME` must be available in the initial user/mount namespace for access to remote configuration, cache, and any bootstrap specific config that lives in `$HOME` (e.g. `.rpmmacros`).

A user can perhaps get around this issue with `newgrp` to select the relevant supplementary group as their effective gid before running a build.

Related to #1550, however, we should perhaps be able to avoid calling `setgroups`. This will allow access to files on the host... though anything except the (fake) `gid=0` group will appear to map to `nobody` in the namespace. This, therefore, is not a complete solution to the lack of supplementary group mapping. You cannot use the group inside the container, other than access existing files on the host through it.

dtrudg:

When we run in `--fakeroot` mode, the `starter` will call `setgroups` and reduce the group membership of the container process to `gid=0` only, inside the fakeroot id-mapped user namespace.

This removes supplementary group membership, and therefore prevents access to files on the host that are restricted by group ownership where the group is a supplementary group of the current user.

If, for example, `$HOME` is accessed via a supplementary group then it cannot be mounted inside the id-mapped user namespace created at the start of a `singularity build --fakeroot`. The build will then fail.

This PR adds a new flag `--no-setgroups` to skip the `setgroups` call in `--fakeroot` mode. This results in the container process *not* having its group membership changed. Host files accessible via host supplementary groups will remain accessible.

This does not make the supplementary groups fully available in the container. Inside the id-mapped user namespace they will be mapped to `nobody`. The output of `id` may be confusing, and we still cannot e.g. `chgrp` a file to one of the user's supplementary groups.

On host...

```
$ groups
dtrudg-sylabs wheel mock docker libvirt
```

In container with `--fakeroot --no-setgroups`...

```
Singularity> groups
root nobody nobody nobody nobody
```

 - Fixes #1748 

### DIFF
--- a/internal/pkg/runtime/engine/config/starter/starter_linux.go
+++ b/internal/pkg/runtime/engine/config/starter/starter_linux.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2022, Sylabs Inc. All rights reserved.
+// Copyright (c) 2018-2023, Sylabs Inc. All rights reserved.
 // This software is licensed under a 3-clause BSD license. Please consult the
 // LICENSE.md file distributed with the sources of this project regarding your
 // rights to use or distribute this software.
@@ -197,6 +197,18 @@ func (c *Config) SetAllowSetgroups(allow bool) {
 		c.config.container.privileges.allowSetgroups = C.true
 	} else {
 		c.config.container.privileges.allowSetgroups = C.false
+	}
+}
+
+// SetNoSetgroups disables the setgroups call for the container process in the
+// starter. Preserves access to files that depends on supplementary groups
+// outside of the user namespace. The supplementary groups will map to 'nobody'
+// inside the container.
+func (c *Config) SetNoSetgroups(noSetgroups bool) {
+	if noSetgroups {
+		c.config.container.privileges.noSetgroups = C.true
+	} else {
+		c.config.container.privileges.noSetgroups = C.false
 	}
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,12 @@
   `SINGULARITY_CONFIGDIR` environment variable.
 - If kernel does not support unprivileged overlays, OCI-mode will attempt to use
   `fuse-overlayfs` and `fusermount` for overlay mounting and unmounting.
+- Added `--no-setgroups` flag for `--fakeroot` builds and run/shell/exec. This
+  prevents the `setgroups` syscall being used on the container process in the
+  fakeroot user namespace. Maintains access from within the user namespace to
+  files on the host that have permissions based on supplementary group
+  membership. Note that supplementary groups are mapped to `nobody` in the
+  container, and `chgrp`, `newgrp`, etc. cannot be used.
 
 ### Bug Fixes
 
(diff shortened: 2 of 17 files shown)


tri-adam (approved):
LGTM

dtrudg:
Thanks @tri-adam, I'll wait for @preminger to review also, before merging.

internal/pkg/runtime/engine/config/starter/starter_linux.go
```diff
@@ -200,6 +200,18 @@ func (c *Config) SetAllowSetgroups(allow bool) {
 	}
 }
 
+// SetNoSetgroups disables the setgroups call for the container process in the
+// starter. Preserves access to files that depends on supplementary groups
+// outside of the user namespace. The supplementary groups will map to 'nobody'
+// inside the container.
+func (c *Config) SetNoSetgroups(allow bool) {
```
preminger, line 207: I found the name `allow` for the argument here a bit hard to grok. Especially since, in this context, passing `true` _prevents_ something from happening (namely, a setgroups call).

preminger (approved):
LGTM, but see small comment on argument naming.

DrDaveD:
I looked into importing this into apptainer, but it turns out that since apptainer/apptainer#915 it already skips calling that setgroups() function by default, so this is not needed.

It's very interesting though that supplementary groups still work inside of a user namespace, they just don't look like they will since the files show up as gid 65534.  I hadn't realized that.  This works in both --fakeroot mode and --userns mode. 

Apptainer also consolidates multiple 65534 groups inside the namespace into one since apptainer/apptainer#1296.

dtrudg:
> It's very interesting though that supplementary groups still work inside of a user namespace, they just don't look like they will since the files show up as gid 65534. I hadn't realized that. This works in both --fakeroot mode and --userns mode.

I'd probably say that they actually aren't really working in the user namespace, since due to the nobody mapping you can't use / become these groups normally inside the container. They are still working outside of the user namespace, where the file permissions are ultimately enforced. The kernel is just still tracking membership through the subgid mapping... so file access on the host still works.

It isn't a complete solution to the supplementary group issues. Try e.g. creating _new_ files from the container and attempting a  chgrp to make them a member of a 'project group' or similar... but it helps a bit.

dtrudg:
An illustration...

```
# I'm in various groups on the host including `testgroup` and `wheel`
$ groups
dtrudg-sylabs wheel mock docker libvirt testgroup

# An existing file on the host is restricted by the group...
$ ls -lah in
-rwxr-x---. 1 root testgroup 7 Jun 26 09:13 in

# Access to existing file now works with --no-setgroups
$ singularity run --fakeroot --no-setgroups docker://alpine
Singularity> cat in
Hello!

# But I can't do anything else in the container with the group... 
# so it's not a solution to many HPC use cases for shared projects etc.
Singularity> chgrp testgroup out
chgrp: unknown group testgroup

# If I try to use a supplementary group that is defined in the container /etc/group, it appears to work...
Singularity> chgrp wheel out

# But back on the host, it's not in `wheel`... it's in my subgid range
$ ls -lah out 
-rw-r--r--. 1 dtrudg-sylabs 165545 0 Jun 26 09:14 out
```

PR #1774, opened 2023-06-14
Enter namespaces earlier in --oci mode. 

### LINKED ISSUE #1773: Move --oci mode bundle setup inside initial user & mount namespace
**Type of issue**

technical debt

**Description of issue**

With the basic overlay work done, we start to have a structure to move toward mounts of images in `--oci` mode, e.g. #1710 

We want to perform images mounts inside the initial root-mapped user namespace / mount namespace, so that they don't pollute the host, don't require setuid FUSE, and for tidier cleanup.

The easiest way to do this is to move the entire oci bundle setup inside the initial user & mount namespace... making almost everything occur inside the namespace. This is somewhat analagous to the way that native mode builds work.

dtrudg:

Before this PR, if we are a non-root user we enter a root-mapped user namespace before calling `singularity oci run` at the end of the `--oci` launcher flow. If we are the root user, we do not enter any namespace. This means that bundle preparation, performed by the launcher, takes place outside of any namespaces.

As a non-root user, we want to perform bundle preparation inside a root-mapped user namespace. This is in order to benefit from fully-unprivileged FUSE operations when we start handling image mounts, plus advantages around cleanup on namespace exit. As a root user, we want to be inside a mount namespace to avoid polluting the host mount namespace with our container-specific mounts.

In this PR:

* As a non-root user, a root-mapped namespace is entered early, before   the oci launcher code executes.
* As a root user, a mount namespace is entered early, before the oci launcher code executes.

We use the `_CONTAINERS_ROOTLESS_xID` environment variables for tracking our UID/GID outside of the root-mapped namespace, since they must be set anyway for `containers/image` to operate properly inside the namespace.  We are using `containers/image` for our OCI image handling.

 - Fixes #1773 

### DIFF
--- a/internal/pkg/runtime/launcher/oci/oci_runc_linux.go
+++ b/internal/pkg/runtime/launcher/oci/oci_runc_linux.go
@@ -14,12 +14,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 
-	"github.com/sylabs/singularity/internal/pkg/buildcfg"
-	fakerootConfig "github.com/sylabs/singularity/internal/pkg/runtime/engine/fakeroot/config"
-	"github.com/sylabs/singularity/internal/pkg/util/starter"
-	"github.com/sylabs/singularity/pkg/runtime/engine/config"
 	"github.com/sylabs/singularity/pkg/sylog"
 )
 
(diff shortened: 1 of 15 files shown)


internal/pkg/runtime/launcher/oci/oci_runc_linux.go
```diff
@@ -197,6 +192,7 @@ func Run(ctx context.Context, containerID, bundlePath, pidFile string, systemdCg
 	}
 
 	runtimeArgs := []string{
+		"--debug",
```
preminger, line 195: Is this here intentionally, or a leftover from local debugging? If the former, it might warrant a code comment to explain.

dtrudg: Sorry - this will go away... I'm going to leave it there until I can test on some older distros etc.

PR #1822, opened 2023-06-28
don't bind-mount /etc passwd & group when persistent writes possible

### LINKED ISSUE #1812: Revamped passwd utility interferes with groupadd and installing packages
**Version of Singularity**

main branch, in particular pr #1597

**Describe the bug**

When attempting to import #1597 into Apptainer, we found that an e2e test that does `apt-get update && apt-get install openssh-client` in an Ubuntu 22.04 fakeroot writable-tmpfs container was failing with an error
```
groupadd: failure while writing changes to /etc/group
addgroup: `/sbin/groupadd -g 101 _ssh' returned error code 10. Exiting.
```
I reproduced it interactively and found that the groupadd function was creating a new temporary file and attempting to rename it to `/etc/group`.  That was failing with a `Device or resource busy` error, because /etc/group is a tmpfs mount after that PR.

**To Reproduce**

```
singularity shell -f --writable-tmpfs docker://ubuntu:22.04
groupadd -g 101 ssh
```

**Expected behavior**

Successful addition of a group

**Actual behavior**

Error message:
```
groupadd: failure while writing changes to /etc/group
```

**OS / Linux Distribution**

I don't think it matters, but I was using Centos 7.

preminger:

If all of the following conditions hold —

- The user inside the container is root
- The root filesystem is writable
- The root filesystem is not a tmpfs

— then writes to `/etc/passwd` or `/etc/group` are possible, and would be expected to persist from build to action stage (run/exec/shell) or across distinct actions (e.g. from one run to the next).

Under this exact combination of conditions, then, we inhibit bind-mounting modified versions of `/etc/passwd` and `/etc/group` from tmpfs storage into the container.

 - Fixes #1812

### DIFF
--- a/e2e/imgbuild/imgbuild.go
+++ b/e2e/imgbuild/imgbuild.go
@@ -2021,5 +2021,6 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
 		"issue 5668":                      c.issue5668,                    // https://github.com/hpcng/singularity/issues/5435
 		"issue 5690":                      c.issue5690,                    // https://github.com/hpcng/singularity/issues/5690
 		"issue 1273":                      c.issue1273,                    // https://github.com/sylabs/singularity/issues/1273
+		"issue 1812":                      c.issue1812,                    // https://github.com/sylabs/singularity/issues/1812
 	}
 }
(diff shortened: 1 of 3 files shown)


dtrudg (approved):
image

PR #1831, opened 2023-06-30
replace build-args Transformer with full-text replacement

preminger:

In #1738, the `transform.Transformer` object used to replace build-args (e.g. `"{{ argname }}"`) to enable templating in definition files (currently in internal/pkg/build/build.go) could encounter _partial input_ scenarios - where the bytes it has read contain the beginning of a build-arg, but it gets clipped due to the block size (e.g. `"{{ argname }"`, `"{{ argna"`, `"{"`, etc.).

The `transform.Transformer` [API](https://pkg.go.dev/golang.org/x/text/transform#Transformer) is not well-suited to facilitate a workaround for this type of problem. That's because `transform.Transformer` is fundamentally built for byte-by-byte transformations, rather than transformations with significantly richer state.

To correct this problem, this PR replaces the previous implementation of build-args templating - which used `transform.NewReader()` to wrap the underlying definition-file Reader in a custom `transform.Transformer` - with an implementation that simply reads in the full text of the definition file, wraps the modified text in a new Reader, and passes that to the code that builds `types.Definition` objects.


dtrudg (commented):
Thanks for considering the incomplete match issue with the transformer. Apologies I also missed that issue in the original implementation.

I'm in two minds on this PR tbh... the general wish to keep things simple has me wanting to drop back to reading everything and doing a simple replace over the entire content in a string or []byte, rather than handle the buffering in somewhat complex logic.

Elsewhere, we are making multiple copies of the stage and full raw def in memory (we set it in the types.Definition for every stage), so there isn't much gain by doing things with a buffered reader here?

If you do feel strongly for the buffered reader approach please could you consider the following for this PR:
 
 1) Implementing some kind of unit test for the new args.Reader that explicilty tests for an incomplete arg in the buffered read. I think you can set bufio.Size and construct a carefully aligned input test case to do that.
 
 2) Reducing the amount of Debug logging. It's quite chatty for something that shouldn't need end-user debugging (since we should have tests that prove it works, and it's not something interacting with the system externally).

internal/pkg/build/args/reader.go
```diff
@@ -0,0 +1,120 @@
...
+	return r
+}
+
+func (r *Reader) Read(p []byte) (n int, err error) {
+	var srcBytes []byte
+	var matches [][]int
+	atRealEOF := false
+	for !atRealEOF {
+		srcBufSize := r.src.Buffered()
+		sylog.Debugf("Underlying bufio.Reader has %d buffered bytes", srcBufSize)
```
dtrudg, line 58: This level of debug message is too chatty for the end-user.

Ideally this Read function is proven working by unit tests. It doesn't interact with the external system, so we shouldn't need this level of info shown to the user.

internal/pkg/build/args/reader.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"io"
+	"regexp"
+
+	"github.com/samber/lo"
+)
+
+var buildArgsRegexp = regexp.MustCompile(`{{\s*(\w+)\s*}}`)
+
+func NewReader(src io.Reader, buildArgsMap map[string]string, defaultArgsMap map[string]string, consumedArgs *[]string) (io.Reader, error) {
```
dtrudg, line 21: Would prefer it if it returned `consumedArgs` as a `[]string`, and the append was done by the caller.

If you don't change this, please make sure a function comment is added here the documents the side-effect on the passed in consumedArgs.

dtrudg (approved):
Generally LGTM - please see comment about side-effect / function comment, but I don't need to re-review if you just add the function comment.

Tempted to ask for a TestNewReader unit test... to make it easier if the implementation is revisited, but will leave that up to you.

internal/pkg/build/args/reader.go
preminger: My thinking here is that in the previous (now scrapped) implementation, the NewReader() constructor did not yet "know" which arguments were consumed; that was (and could only be) calculated in the course of Reader.Read(). The way this implementation works, NewReader() actually returns a plain-vanilla bytes.Reader, so the Read() method is "naive", and the knowledge of which arguments got consumed is already present in the constructor. This means that before, we got the consumed args info as part of the return value of Read(), and now, if we did it as a return value, we'd get it as part of the return value for the NewReader() constructor.

This implementation-dependence strikes me as a bad sign, from a software-design perspective. So I've gone with the current approach; say we at some point decide to switch back to an implementation where Read() is the one doing the actual replacement work - we could then have NewReader() just store the consumedArgs pointer in a private struct member, and have (our new, custom) Read() populate it, with no change to the API.

So moving away from that implementation-dependence was my motivation here.

I will of course add the function comment.

preminger:
> Tempted to ask for a TestNewReader unit test... to make it easier if the implementation is revisited, but will leave that up to you.

Beyond this...? https://github.com/sylabs/singularity/blob/5c926ef4a74628c174626286cc6af07594bdc27f/internal/pkg/build/args/args_test.go#L93

(Given that Read() is now the run-of-the-mill Read() method of bytes.Reader, all that our TestReader() is testing is NewReader()...)

dtrudg:
> > Tempted to ask for a TestNewReader unit test... to make it easier if the implementation is revisited, but will leave that up to you.
> 
> Beyond this...?
> 
> https://github.com/sylabs/singularity/blob/5c926ef4a74628c174626286cc6af07594bdc27f/internal/pkg/build/args/args_test.go#L93
> 
> (Given that Read() is now the run-of-the-mill Read() method of bytes.Reader, all that our TestReader() is testing is NewReader()...)

Yeah... ignore this 🤦  - I was looking for a `reader_test.go` and didn't clock that bit. Sorry!

dtrudg (approved):

Issue #1861, opened 2023-07-07
oci: Auto-detect oci-sif and handle without `oci-sif:` prefix


dtrudg:
Currently, `singularity run/shell/exec` does not require a prefix to be specified when running a native SIF image.

In `--oci` mode via #1838 , the `oci-sif:` prefix must currently be used to run an oci-sif image.

There should be some auto-detection of the image type, so that oci-sif can be run without the prefix, to match the native singularity runtime:

```
singularity --oci run my-oci-sif.sif
```

PR #1886, opened 2023-07-13
pkg/image: Add minimal oci-sif type

dtrudg:
Add a minimal oci-sif image type to pkg/image, that does not implement any functionality other than oci-sif detection.

Ensure oci-sif detected as oci-sif, and are not detected as the existing sif type.

Modify the native launcher to return a more friendly error if a user attempts to run an oci-sif with it.

Prep work for #1861

Example...

```
$ singularity run alpine_latest.sif 
FATAL:   While checking image: native runtime does not support OCI-SIF images, use --oci mode
```

### DIFF
--- a/pkg/image/ocisif.go
+++ b/pkg/image/ocisif.go
@@ -1,0 +1,62 @@
+// Copyright (c) 2018-2023, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package image
+
+import (
+	"bytes"
+	"os"
+
+	"github.com/sylabs/sif/v2/pkg/sif"
+)
+
+type ociSifFormat struct{}
+
+// initializer performs minimal detection of oci-sif images only.
+// It does not populate any information except img.Type.
+func (f *ociSifFormat) initializer(img *Image, fi os.FileInfo) error {
+	if fi.IsDir() {
+		return debugError("not an oci-sif file image")
+	}
+	b := make([]byte, bufferSize)
+	if n, err := img.File.Read(b); err != nil || n != bufferSize {
+		return debugErrorf("can't read first %d bytes: %v", bufferSize, err)
+	}
+	if !bytes.Contains(b, []byte("SIF_MAGIC")) {
+		return debugError("SIF magic not found")
+	}
+
+	flag := os.O_RDONLY
+	if img.Writable {
+		flag = os.O_RDWR
+	}
+
+	// Load the SIF file
+	fimg, err := sif.LoadContainer(img.File,
+		sif.OptLoadWithFlag(flag),
+		sif.OptLoadWithCloseOnUnload(false),
+	)
+	if err != nil {
+		return err
+	}
+	defer fimg.UnloadContainer()
+
+	// It's a SIF, but an OCI-SIF.
+	if _, err := fimg.GetDescriptor(sif.WithDataType(sif.DataOCIRootIndex)); err != nil {
+		return debugErrorf("image is not an OCI-SIF")
+	}
+
+	img.Type = OCISIF
+
+	return nil
+}
+
+func (f *ociSifFormat) openMode(writable bool) int {
+	return os.O_RDONLY
+}
+
+func (f *ociSifFormat) lock(img *Image) error {
+	return nil
+}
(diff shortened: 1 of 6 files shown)


pkg/image/ocisif.go
```diff
@@ -0,0 +1,62 @@
...
+type ociSifFormat struct{}
+
+// initializer performs minimal detection of oci-sif images only.
+// It does not populate any information except img.Type.
+func (f *ociSifFormat) initializer(img *Image, fi os.FileInfo) error {
+	if fi.IsDir() {
+		return debugError("not an oci-sif file image")
+	}
+	b := make([]byte, bufferSize)
+	if n, err := img.File.Read(b); err != nil || n != bufferSize {
```
preminger, line 24: I see that `bufferSize` is defined in pkg/image/image.go:55 to be 2048. Is it categorically impossible to have valid image files (be they oci-sif, or some other type for that matter) that are smaller than that?

pkg/image/sif.go
```diff
@@ -78,6 +78,11 @@ func (f *sifFormat) initializer(img *Image, fi os.FileInfo) error {
 	}
 	defer fimg.UnloadContainer()
 
+	// It's a SIF, but an OCI-SIF.
+	if _, err := fimg.GetDescriptor(sif.WithDataType(sif.DataOCIRootIndex)); err == nil {
+		return debugErrorf("image is an OCI-SIF")
```
preminger, line 83: Am I correct in understanding that this is a circumstance that shouldn't arise (i.e., there is code elsewhere that should make sure we don't reach this point if we're dealing with an OCI-SIF image)? And if so, wouldn't it be more accurate to characterize this circumstance as an "internal error" than a "debug error"...?

preminger (approved):
LGTM, but see two comments.

pkg/image/ocisif.go
dtrudg: Yes - at present an empty SIF is 32KiB.

A blank SIF will consist of a global header, followed by an area reserved for up to n (where n currently = 48) descriptors, prior to the variable sized data area.

The descriptor area is aligned to start at a 4096 byte boundary, so even if n were smaller, the SIF file is at least 4096 bytes.

It may be the case that the sif package should export a constant, but this value has been used here since the inception of SIF support, I believe.

pkg/image/sif.go
dtrudg: No - this error is expected to be encountered. When Singularity's native runtime related code is provided an image file, it uses pkg/image to try and identify/load the image... and this attempts to call the initializer of each pkg/image type in turn. We can see this in the debug trace, for when I try to run an `oci-sif` with the native runtime:

```
$ singularity -d run  alpine_latest.sif 
DEBUG   [U=1001,P=1731880] persistentPreRun()            Singularity version: 3.11.0+462-g64fd3bd95
DEBUG   [U=1001,P=1731880] persistentPreRun()            Parsing configuration file /usr/local/etc/singularity/singularity.conf
DEBUG   [U=1001,P=1731880] maybeReExec()                 Checking whether to re-exec
DEBUG   [U=1001,P=1731880] handleConfDir()               /home/dtrudg-sylabs/.singularity already exists. Not creating.
DEBUG   [U=1001,P=1731880] handleRemoteConf()            Ensuring file permission of 0600 on /home/dtrudg-sylabs/.singularity/remote.yaml
DEBUG   [U=1001,P=1731880] launchContainer()             Using native runtime launcher.
DEBUG   [U=1001,P=1731880] setUmask()                    Saving umask 0022 for propagation into container
DEBUG   [U=1001,P=1731880] Init()                        Image format detection
DEBUG   [U=1001,P=1731880] Init()                        Check for sandbox image format
DEBUG   [U=1001,P=1731880] Init()                        sandbox format initializer returned: not a directory image
DEBUG   [U=1001,P=1731880] Init()                        Check for sif image format
DEBUG   [U=1001,P=1731880] Init()                        sif format initializer returned: image is an OCI-SIF
DEBUG   [U=1001,P=1731880] Init()                        Check for ocisif image format
DEBUG   [U=1001,P=1731880] Init()                        ocisif image format detected
FATAL   [U=1001,P=1731880] Exec()                        While checking image: native runtime does not support OCI-SIF images, use --oci mode
``` 

Note how it checks for a sandbox/dir first... then a SIF (non-OCI). Both of those initializers have to return a `debugError` to allow fall-through to trying the next initializer.

The initializer for ocsif succeeds... so the returned `img.Type` is `OCISIF`, and the rejection of execution is based on that.

The intent of the PR is really to modify `pkg/image` so it recognises `OCISIF` as a separate image type, and doesn't identify it as a SIF.

PR #1902, opened 2023-07-17
oci: default to /.singularity.d/runscript for SIF bundle

### LINKED ISSUE #1898: oci: change singularity-sif bundle CMD to runscript
When a singularity SIF is mounted to an OCI bundle via `singularity oci mount` or through code, the bundle runtime config defaults to executing `/bin/sh`.

This does not reflect the intended behaviour of a Singularity container, which is to execute the `/.singularity.d/runscript`.

An oci bundle created from a singularity SIF should execute the runscript by default.

dtrudg:

When a bundle is created from a singularity SIF (e.g. via `singularity oci mount`), the default process is now the container runscript.

Previously the process specified in the OCI runtime config was `/bin/sh`, resulting in the default behaviour of the bundle not matching execution under singularity.

 - Fixes #1898

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@
 - Lookup and store user/group information in stage one prior to entering any
   namespaces to fix issue with winbind not correctly lookup user/group
   information when using user namespace.
+- The default process for an OCI bundle created with `singualrity oci mount` is now
+  `/singularity.d/runscript` instead of `/bin/sh`.
 
 ### New Features & Functionality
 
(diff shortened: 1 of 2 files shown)


dtrudg:
While working through #1906 it became clear that this is not a good idea.

With no sensible direct mapping of runscript/startscript behaviour into CMD/ENTRYPOINT it doesn't really make sense to pick an approach at this level of the code. The caller has to decide what to do, and a simple `/bin/sh` is an appropriate fall-back.

PR #1920, opened 2023-07-20
`remote add` sets new endpoint as default, suppressible via flag

### LINKED ISSUE #1895: automatically select argument to `remote add` as current remote
When adding a remote via `singularity remote add <remote_name> <remote_uri>`, the newly-added remote should be selected as the _current_ remote by default.

A flag (`--notcurrent`) should be added to suppress this behavior if needed.

ETA: The flag should be `--not-current`

Discussion: #1639

preminger:

When adding a remote via `singularity remote add <remote_name> <remote_uri>`, the newly-added remote will be selected as the _current_ remote endpoint by default.

This can be suppressed (and thus, returned to the previous behavior of `remote add`) by supplying the `-n`/`--no-default` flag to `remote add`.

 - Fixes #1895


cmd/internal/cli/remote.go
```diff
@@ -163,6 +164,16 @@ var remoteAddInsecureFlag = cmdline.Flag{
 	Usage:        "allow connection to an insecure http remote",
 }
 
+// -i|--insecure
+var remoteAddNotDefaultFlag = cmdline.Flag{
+	ID:           "remoteAddNotDefaultFlag",
+	Value:        &remoteAddNotDefault,
+	DefaultValue: false,
+	Name:         "not-default",
```
dtrudg, line 172: I'm leaning toward `no-default`, even if it's not so gramatically correct, because we have a lot of flags with the general pattern `--no-XXXXX`.

If you feel strongly about it being `not-default` then it's probably a nice idea to have `no-default` working also.

dtrudg (changes requested):
Needs e2e testing in e2e/remote/remote.go

The unit tests against `remoteAdd()` are using a positive 'makeDefault'.

Since the CLI behaviour is the opposite, a negative 'not- default', an e2e test is warranted.

The e2e remote tests are poorly structured, and it might be a bit of a pain to add these checks, but we should take this opportunity to improve the coverage of the CLI.

preminger:
> The unit tests against `remoteAdd()` are using a positive 'makeDefault'.

Wait, are we looking at the same thing here? I doubled the set of tests in TestRemoteAdd() in internal/app/singularity/remote_add_test.go - instead of 30, there are now 60 - and the two blocks differ in precisely the value of `tt.makeDefault` they have.

(I think it's possible that GitHub didn't show this because of its "large commits are hidden by default" thing.)

Plus, TestRemoteRemove() in internal/app/singularity/remote_remove_test.go also tests both settings of `makeDefault` now.

dtrudg (approved):

PR #1926, opened 2023-07-24
pull: Pull previously pushed OCI-SIF from OCI registry (docker://)

### LINKED ISSUE #1915: Implement pull of oci-sif from OCI registries (docker://)
Inverse of #1914 

An OCI-SIF image that was pushed to an OCI registry will be present in the registry as a single-layer, squashfs layer-format, OCI image.

It should be possible to pull this type of image back down into an OCI-SIF file directly, without any mutation/conversion of the remote image or its blobs.

dtrudg:

If `singularity pull --oci` is used against an OCI registry `docker://` reference, and the reference is a single-layer squashfs image, pull it into an OCI-SIF without any mutation.

This is the mirror-image operation of #1917, allowing OCI-SIF images pushed with the code in #1917 to be pulled.

Note that the pull e2e tests have been re-arranged / re-named. This is to keep all the `library://` `oras://` `docker://` tests next to eachother.... and in an attempt to make it more obvious what is testing what.

 - Closes #1915 


internal/pkg/client/oci/ocisif.go
```diff
@@ -168,10 +195,7 @@ func layoutToOciSif(layoutDir string, digest v1.Hash, imageDest, workDir string)
 		return fmt.Errorf("while replacing layers: %w", err)
 	}
 
-	sylog.Infof("Writing oci-sif image...")
-	if err := lp.ReplaceImage(img, match.Digests(digest)); err != nil {
```
dtrudg, line 172: Note that this deletion means that the squashed, mutated image is only written into the final SIF, and not back into the temporary layout.

I had left it writing into the temporary layout for some speculative reason that escapes me, previously... but confirmed it is not needed in this PR... which should be the final significant change for pkg/client/oci/ocisif

PR #1927, opened 2023-07-24
library:// oci-sif push/pull

### LINKED ISSUE #1916: Implement push/pull of oci-sif to library:// backing OCI registry

dtrudg:
When `singularity push` is provided an OCI-SIF image file and a `library://` URI, push as a single layer squashfs OCI image to the OCI registry that backs the library service. Existing push behaviour for native SIF through the library API remains.

When `singularity pull --oci` is provided a `library://` URI with a path that points to a single layer squashfs OCI image in the backing OCI registry, pull to a local OCI-SIF image.

When `singularity pull` is provided a `library://` URI and no `--oci` flag, try a native SIF library API pull first. If that fails and the error indicates the image is an OCI image then try an OCI-SIF pull.

The push/pull process involves discovering the URI of the backing OCI registry, retrieving an authentication token, and translating `library://` URIs to OCI registry URIs.

Because a `pull` of an standard OCI image to a native SIF is implemented as a build, and the build code imports from
`pkg/client/*`, we have to do some gymnastics to avoid introducing import loops:

* Code dealing with oci image operations used by both builds and the oci client package has been moved to `internal/pkg/ociimage`.

* Low-level push/pull between OCI-SIF and OCI registries has been placed in `internal/pkg/client/ocisif` rather than `oci`.

Also tidy up the `--oci` flag CLI visibility:

* Un-hide the `--oci` flag on the `pull` command.
    
* Warn where it has no effect (oras/shub/http[s] sources are currently direct downloads without enforcing `--oci` gives an OCI-SIF.

* Explain the behaviour in the `pull --help` output.

Closes #1916

### DIFF
--- a/internal/pkg/client/library/push.go
+++ b/internal/pkg/client/library/push.go
@@ -16,27 +16,51 @@ import (
 	scslibrary "github.com/sylabs/scs-library-client/client"
 	"github.com/sylabs/sif/v2/pkg/sif"
 	"github.com/sylabs/singularity/internal/pkg/client"
+	"github.com/sylabs/singularity/internal/pkg/client/ocisif"
+	"github.com/sylabs/singularity/internal/pkg/remote/endpoint"
 	"github.com/sylabs/singularity/pkg/sylog"
 	"golang.org/x/term"
 )
 
+// PushOptions provides options/configuration that determine the behavior of a
+// push to the library.
+type PushOptions struct {
+	// Description sets the optional description for an image pushed via the
+	// library's own API.
+	Description string
+	// Endpoint is the active remote endpoint, against which the OCI registry
+	// backing the library can be discovered.
+	Endpoint *endpoint.Config
+	// LibraryConfig configures operations against the library using its native
+	// API, via sylabs/scs-library-client.
+	LibraryConfig *scslibrary.Config
+}
+
 // Push will upload an image file to the library.
-// Returns the upload completion response on success, containing container path and quota usage.
-func Push(ctx context.Context, sourceFile string, destRef *scslibrary.Ref, desc string, libraryConfig *scslibrary.Config) (uploadResponse *scslibrary.UploadImageComplete, err error) {
-	fi, err := os.Stat(sourceFile)
+// Returns the upload completion response on success, if available, containing
+// container path and quota usage for v1 libraries.
+func Push(ctx context.Context, sourceFile string, destRef *scslibrary.Ref, opts PushOptions) (uploadResponse *scslibrary.UploadImageComplete, err error) {
+	f, err := sif.LoadContainerFromPath(sourceFile, sif.OptLoadWithFlag(os.O_RDONLY))
 	if err != nil {
-		if os.IsNotExist(err) {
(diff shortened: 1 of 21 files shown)


preminger (commented):
@EmmEff: I've looked through all the parts of this PR where I feel my knowledge of the relevant code & what it interacts with are at a sufficient level for a contentful review. I think we need your eyes on this, too, before we can consider it thoroughly reviewed, though.

internal/pkg/client/library/push.go
```diff
@@ -52,18 +76,28 @@ func Push(ctx context.Context, sourceFile string, destRef *scslibrary.Ref, desc
...
-		if err == nil && resp != nil && progressBar == nil {
-			sylog.Infof("Uploaded %d bytes in %v\n", fi.Size(), time.Since(t))
+		if err == nil && progressBar == nil {
+			sylog.Infof("Uploaded %d bytes in %v\n", fSize, time.Since(t))
 		}
 	}(time.Now())
 
+	resp, err := libraryClient.UploadImage(ctx, f, destRef.Path, arch, destRef.Tags, opts.Description, progressBar)
+
 	return resp, err
```
EmmEff, line 101:
```suggestion
	return libraryClient.UploadImage(ctx, f, destRef.Path, arch, destRef.Tags, opts.Description, progressBar)
```

EmmEff (approved):
One minor nit

internal/pkg/client/ocisif/ocisif.go
```diff
@@ -202,12 +228,13 @@ func convertLayoutToOciSif(layoutDir string, digest v1.Hash, imageDest, workDir
...
 }
 
-// pushOCISIF pushes a single image from an OCI SIF to the OCI registry destination ref.
-func pushOCISIF(ctx context.Context, sourceFile, dest string, ociAuth *ocitypes.DockerAuthConfig) error {
-	dest = strings.TrimPrefix(dest, "//")
-	ref, err := name.ParseReference(dest)
+// PushOCISIF pushes a single image from sourceFile to the OCI registry destRef.
+func PushOCISIF(ctx context.Context, sourceFile, destRef string, ociAuth *ocitypes.DockerAuthConfig) error {
+	destRef = strings.TrimPrefix(destRef, "docker://")
+	destRef = strings.TrimPrefix(destRef, "//")
```
EmmEff, line 234: I am not understanding how `destRef` could  be either `docker:////` or `//<something>`. Is line 234 necessary?

dtrudg: Unfortunately there are differences between how higher-level code, across builds, the `pull` CLI, and the `actions` CLI send references to clients. This is a defence against that.

There has to be stripping of `docker://` to handle when the full ref is sent, and `//` for the case where the caller strips the `docker:` part, but not the following `//`.

Same for oras:

image

I really want to tidy this up, but there are many paths here to rationalise... so I'll tackle it separately. Ultimately the aim is to switch to using only google/go-containerregistry, and remove the remaining duplicated code between build / pull / actions which complicates all of this.

PR #1929, opened 2023-07-24
remote: Make registryUri available for library

dtrudg:

Retrieve the `registryUri` value from the library service config for SCS/Enterprise remotes.

The concept of a service in the remote code is rather overloaded and inflexible. Add an internal use map that we can pull the `registryUri` into and out of. Hide it behind a single `RegistryURI` public fn.

Not the tidiest, but it works pending major overhaul of the remote / endpoint code.

### DIFF
--- a/internal/pkg/remote/endpoint/service.go
+++ b/internal/pkg/remote/endpoint/service.go
@@ -1,5 +1,5 @@
 // Copyright (c) 2020, Control Command Inc. All rights reserved.
-// Copyright (c) 2019-2022, Sylabs Inc. All rights reserved.
+// Copyright (c) 2019-2023, Sylabs Inc. All rights reserved.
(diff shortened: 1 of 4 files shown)


internal/pkg/remote/endpoint/service.go
```diff
@@ -109,6 +121,16 @@ func (s *service) Status() (version string, err error) {
 	return vRes.Version, nil
 }
 
+// configVal returns the value of the specified key (if present), in the
+// service's additional known configuration.
+func (s *service) configVal(key string) string {
+	val, ok := s.configMap[key]
```
preminger, line 127: Not sure this pattern is necessary here - doesn't a `map[string]string` already return an empty string upon lookup with a nonexistent key...?

preminger (changes requested):

preminger (approved):

PR #1942, opened 2023-07-26
Add support for %appstart def-file section and `--app` option to `instance start`

### LINKED ISSUE #1346: Cherry pick SCIF appstart addition (apptainer 1125)
**Type of issue**

Other

**Description of issue**

There is a proposal in apptainer to add an `%appstart` section to handling of SCIF apps. I have brought up that this needs to be considered upstream, in the SCIF specification first.

Track the progress / implementation of this: https://github.com/apptainer/apptainer/issues/1118

preminger:

This is a pick of https://github.com/apptainer/apptainer/pull/1125

Adds support for the `%appstart` section in singularity definition files, and an `--app` option in the `instance start` command which will run the commands in that section (if present) upon starting the named instance.

Example:

```
singularity instance start --app tail tailer.sif mytailer
```

Contents of corresponding definition file:

```
%appstart tail
    tail -f /dev/null
```

In this example, an instance called `mytailer` will be created which will tail `/dev/null` until terminated.

Stopping the instance is the exact same as stopping any other singularity instance. The app name is no longer necessary in the `instance stop` command; only the instance name is required:

```
singularity instance stop mytailer
```

 - Fixes #1346


dtrudg (approved):
Cherry-pick looks good to me, but I don't think the original PR includes an e2e test that actually ensures that `instance start --app xxxx` really does the right thing?

Could you open a techdebt issue for that? No rush on this... it does appear to work properly at present :-)

PR #1945, opened 2023-07-26
WIP: deps: Bundle sqfstar, else search sqfstar then tar2sqfs on PATH

### LINKED ISSUE #1941: Bundle recent `sqfstar` from the squashfs-tools project
Creation of OCI-SIF images using sylabs/oci-tools requires either `sqfstar` (from squashfs-tools) or `tar2sqfs` (from squashfs-tools-ng).

While all distros provide squashfs-tools, some (e.g. EL) don't include the `sqfstar` binary.

Although squashfs-tools-ng is packaged for many distributions, it is not for EL.

To work around this, bundle `sqfstar` in a similar way as `squashfuse_ll` was provided in #1937 

There must be an ability to build without a bundles `sqfstar` and fall back to a copy on `PATH`.

We are going with `sqfstar` at this point, because squashfs-tools currently has more regular release activity than `-ng`.

preminger:

Creation of OCI-SIF images using sylabs/oci-tools requires either `sqfstar` (from squashfs-tools) or `tar2sqfs` (from squashfs-tools-ng).

While all distros provide squashfs-tools, some (e.g. EL) don't include the `sqfstar` binary.

Although squashfs-tools-ng is packaged for many distributions, it is not for EL.

Bundle the latest squashfs-tools and build `sqfstar` unless `--without-sqfstar` is specified.

In the case that `--without-sqfstar` is specified, prefer `sqfstar` from `PATH`, falling back to `tar2sqfs`.

 - Fixes #1941

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "third_party/squashfuse"]
 	path = third_party/squashfuse
 	url = https://github.com/vasi/squashfuse
+[submodule "third_party/squashfs-tools"]
+	path = third_party/squashfs-tools
+	url = https://github.com/plougher/squashfs-tools
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,9 @@
   submodule unless `--without-squashfuse` is specified as an argument to
   `mconfig`. When built with `--without-squashfuse`, `squashfuse_ll` or
   `squashfuse` will be located on `PATH`. Version 0.2.0 or later is required.
+- Singularity uses `sqfstar`, which is built from a git submodule unless
+  `--without-sqfstar` is specified as an argument to `mconfig`. When built with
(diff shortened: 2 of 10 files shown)


dtrudg (changes requested):
The code changes look good to me. This is being held as 'Request changes' to block merging due to the GPL license of squashfs-tools.

We are:

- Bringing the code in via git submodule rather than importing it into our repository.
- Compiling `mksquashfs/sqfstar` as a standalone binary only, with no linking (static or dynamic) to our own code... interaction being by CLI, just the same as when the OS provides the tool.

This suggests to me that we should not have an issue with GPL applying to singularity code, but a more authoratative opinion on this is required.

Relevant FSF GPL FAQ item: https://www.gnu.org/licenses/old-licenses/gpl-2.0-faq.en.html#TOCMereAggregation

dtrudg:
Due to licensing concerns, we won't be directly providing `tar2sqfs` or `sqfstar` via this repository, at this time.

PR #1947, opened 2023-07-26
deps: grab newer version of squashfuse with fixed .gitignore

preminger:

https://github.com/vasi/squashfuse/pull/102 adds a few entires that were missing in the project's `.gitignore`. This PR grabs a version with this fix to serve as the bundled `squashfuse` submodule.


dtrudg (approved):
Generally, because this is just a fix to a git status annoyance (and not a critical bug in using singularity), we wouldn't want to bump from a release version of a third party dependency to an unreleased development commit.

In this case, however, the additional stuff that is being brought in lands multi-threaded squashfuse_ll by default... and we are going to need to compile that multi-threaded anyway at some point.

The additional changes look okay to me, and it appears a new release will happen soon.

PR #1964, opened 2023-08-01
Revise CWD creation / symlink handling in native mode (apptainer 1365,1374)

### LINKED ISSUE #1726: Apptainer potential cherry-picks (2023-06-01)
The following should be considered for cherry-picking from Apptainer:

## Straightforward

~~**RFE: Templating Support for Definition File - https://github.com/apptainer/apptainer/pull/1356**
Provides frequently requested definition file templating / build-args. Flags adopted match docker / suggestions in comments, so should be un-controversial for 4.0~~

~~**fix: force casting to uint64 - https://github.com/apptainer/apptainer/pull/1347**
32-bit build fix for instance stats change that originated in CE. Bugfix for main and release-3.11.~~

~~**Add xino=on for writable overlay mount points - https://github.com/apptainer/apptainer/pull/1278/files**
Overlay related fix for inode inconsistency after cache flushing. Bugfix for main and release-3.11.~~

~~**Lookup and store user/group information in stage one prior to creating namespaces - https://github.com/apptainer/apptainer/pull/1368**
Fixes an issue with user lookup on (rare) environments pulling user info from winbind. Bugfix for main and release-3.11.~~

~~**Fix seccomp filters to allow mknod/mknodat syscalls to create pipe/socket** - https://github.com/apptainer/apptainer/pull/1451~~

## Need investigation / discussion

~~**Sort all mount points for a mount tag by shortest destination path first - https://github.com/apptainer/apptainer/pull/1363**
Sorts mount points by depth before mounting, so that a user bind doesn't override a previous bind path. I'm not sure about this one. The semantics of overlapping binds need to be examined carefully. Order does matter with binds, and it's not out of the realms of possibility that there's a situation where we do want to mask a previous bind. At minimum this would be called out as a behaviour change, and for 4.0 only.~~ Not picking this. I remain unconvinced that there aren't valid situations where a CLI bind _should_ be able override binds that come before it.

**Create current working directory in container - https://github.com/apptainer/apptainer/pull/1374**
Tries to revert to >=3.0 <=3.6 behaviour of always creating cwd in container. 3.6 previously reverted to 2.x behaviour of not doing this. Do we want to follow? At minimum this would be called out as a behaviour change, and for 4.0 only.

**Handle CWD symlink schizophrenia between host/container - https://github.com/apptainer/apptainer/pull/1365**
Handle the case where cwd on host and container both contain symlinks, but resolve differently. Don't mount cwd in container in this case, if it already exists. I think this is probably a good one to pick as a bug fix... but needs investigation into whether it's at all tied up with the 2 previous PRs for sensible overall behaviour.

dtrudg:

Cherry-pick two Apptainer PRs:

Create current working directory in container
https://github.com/apptainer/apptainer/pull/1374

fix: Handle CWD symlink differences between host/container
https://github.com/apptainer/apptainer/pull/1365

The effect of these cherry-picks is that we will now create the host CWD path in the container, if it is not already present, so that it can be mounted & entered in more cases.

However, if the CWD exists on the container and inside the host with a path containing symlinks to different locations, we will not perform the mount. There is no other sensible choice in this case.

 - Fixes #1726 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@
 - When `singularity registry login` is used to login to a public OCI registry,
   the password is not verified at login time.
 - Improved the clarity of `singularity key list` output.
+- Do not mount current working directory when its path in the container and on
+  the host contain symlinks to different locations. 
 
 ### New Features & Functionality
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,11 @@
   the password is not verified at login time.
 - Improved the clarity of `singularity key list` output.
 - Do not mount current working directory when its path in the container and on
-  the host contain symlinks to different locations. 
+  the host contain symlinks to different locations.
+- Create current working directory in container when it doesn't exist, so that
+  it can be entered. You must now specify `--no-mount home,cwd` instead of just
+  `--no-mount home` to avoid mounting from `$HOME` if you run `singularity` from
+  inside `$HOME`.
 
 ### New Features & Functionality
 
--- a/e2e/actions/actions.go
+++ b/e2e/actions/actions.go
@@ -2332,6 +2332,17 @@ func (c actionTests) actionNoMount(t *testing.T) {
 			testDefault:   true,
 			testContained: true,
 			exit:          0,
+			// run from / to avoid /home mount via CWD mount
+			cwd: "/",
+		},
+		{
+			// test by excluding both home and cwd without chdir
+			name:          "home,cwd",
+			noMount:       "home,cwd",
+			noMatch:       "on /home",
+			testDefault:   true,
+			testContained: true,
+			exit:          0,
 		},
 		{
 			// /srv is an LSB directory we should be able to rely on for our CWD test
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -55,6 +55,11 @@
 - When `singularity registry login` is used to login to a public OCI registry,
   the password is not verified at login time.
 - Improved the clarity of `singularity key list` output.
+- Do not mount current working directory when its path in the container and on
+  the host contain symlinks to different locations.
+- Create current working directory in container when it doesn't exist, so that
+  it can be entered. If a container is run from inside `$HOME`, `--no-mount
```
preminger, line 61: I find this changelog message a bit unclear. In particular, it reads almost as if the user has to *supply* the `--no-mount` flag, as opposed to it being effectively inferred...

dtrudg: You must now specify `--no-mount home,cwd` instead of just `--no-mount home` to avoid mounting from `$HOME` if you run singularity from inside `$HOME`.

preminger (approved):
LGTM!

PR #1985, opened 2023-08-03
no-compat: read-only unless --writable-tmpfs requested

### LINKED ISSUE #1980: --no-compat: disable writable-tmpfs by default
When `--oci` mode is run with `--no-compat`, the container should be read-only with no writable-tmpfs, by default.

Using `--writable-tmpfs` with `--no-compat` should re-enable the writable-tmpfs.

dtrudg:
 
When running in OCI-mode with `--no-compat`, the container rootfs should be read-only by default, as in native mode.
    
Allow `--writable-tmpfs` to re-enable the writable tmpfs behaviour.

 - Fixes #1980 

### DIFF
--- a/internal/pkg/runtime/launcher/oci/launcher_linux.go
+++ b/internal/pkg/runtime/launcher/oci/launcher_linux.go
@@ -90,6 +90,13 @@ func NewLauncher(opts ...launcher.Option) (*Launcher, error) {
 		return nil, err
 	}
 
+	// We are emulating native mode `--compat`, so  we provide a user-writable
+	// tmpfs by default, unless `--no-compat` was requested without
+	// `--writable-tmpfs`.
+	if !lo.NoCompat || lo.WritableTmpfs {
+		lo.WritableTmpfs = true
+	}
+
 	return &Launcher{
 		cfg:             lo,
 		singularityConf: c,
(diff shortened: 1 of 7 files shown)


internal/pkg/runtime/launcher/oci/launcher_linux.go
```diff
@@ -90,6 +90,13 @@ func NewLauncher(opts ...launcher.Option) (*Launcher, error) {
 		return nil, err
 	}
 
+	// We are emulating native mode `--compat`, so  we provide a user-writable
+	// tmpfs by default, unless `--no-compat` was requested without
+	// `--writable-tmpfs`.
+	if !lo.NoCompat || lo.WritableTmpfs {
```
preminger, line 96: Coding style question: this line only works by virtue of operator precedence rules. Of course, there are certain cases where we wouldn't consider that a barrier to readability (`x+y*z`). My question is whether the precedence of unary negation over binary disjunction is also in that class, or whether wrapping `!lo.NoCompat` in parentheses would be preferable in such cases.

dtrudg: Generally you wouldn't wrap with braces for a `!`. Note that `gofmt` puts the `!` right against the `lo.NoCompat`, but spaces around the `||` to clarify precedence.

FWIW - I'd be more inclined to use parentheses for (x+y*z)

PR #1991, opened 2023-08-04
oci: PATH handling for native SIF images

### LINKED ISSUE #1900: oci: PATH environment handling for singularity SIF
When a singularity SIF (non oci-sif) is executed in `--oci` mode, it should pick up environment in the same manner as with the native runtime.

This means that `PATH` / `APPEND_PATH` / `PREPEND_PATH` specified via `SINGULARITYENV_` / `--env` / `--env-file` should work as in native mode.

dtrudg:

Pass the correct env vars into native SIF images to handle SINGULARITYENV_PATH / SINGULARITYENV_APPEND_PATH / SINGULARITYENV_PREPEND_PATH. These are processed inside the image, by a /.singularity.d/env script.

Don't modify LD_LIBRARY_PATH for native SIF images, as this is handled by a /.singularity.d/env script.

Expand tests for OCI-mode env var handling to cover these circumstances.

 - Fixes #1900

### DIFF
--- a/internal/pkg/runtime/launcher/oci/process_linux.go
+++ b/internal/pkg/runtime/launcher/oci/process_linux.go
@@ -99,7 +99,7 @@ func (l *Launcher) getProcess(ctx context.Context, imgSpec imgspecv1.Image, bund
 		Args:            args,
 		Capabilities:    caps,
 		Cwd:             cwd,
-		Env:             getProcessEnv(imgSpec, rtEnv),
+		Env:             getProcessEnv(imgSpec, rtEnv, l.nativeSIF),
 		NoNewPrivileges: noNewPrivs,
 		User:            u,
 		Terminal:        getProcessTerminal(),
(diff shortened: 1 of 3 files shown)


internal/pkg/runtime/launcher/oci/process_linux.go
```diff
@@ -286,6 +289,26 @@ func getProcessEnv(imageSpec imgspecv1.Image, runtimeEnv map[string]string) []st
...
+		g.AddProcessEnv("LD_LIBRARY_PATH", ldLibraryPath)
+	}
+
+	return g.Config.Process.Env
+}
+
+func setOCIPath(g *generate.Generator, prependPath, path, appendPath string) {
 	// Compute and set optionally APPEND-ed / PREPEND-ed PATH.
 	if appendPath != "" {
 		path = path + ":" + appendPath
```
preminger, line 314: Just making sure: we don't want to do the same `strings.TrimPrefix(..., ":")` here that we do in line 302 of the same file? After all, if `path` being empty is something we're bothering checking for, it stands to reason (I think...?) that `path` can be empty also in a circumstance where `appendPath` is not empty (and therefore the concatenation will occur).

preminger (changes requested):

internal/pkg/runtime/launcher/oci/process_linux.go
dtrudg: Yeah, could do that for neatness and consistency. It doesn't have any practical effect, and PATH is very unlikely to be "" but I'll make the change.

dtrudg:
Via #1992

PR #1992, opened 2023-08-04
oci: allow override of native SIF container env var

### LINKED ISSUE #1990: oci: native SIF environment override
When a singularity SIF (non oci-sif) is executed in --oci mode, it should pick up environment in the same manner as with the native runtime.

In native SIF images, image-specific environment is set by executing `/.singularity.d/env` scripts.

Currently, it is not possible to override an env var set by these scripts using `SINGULARITYENV_` / `--env` / `--env-file`.

Override should work, as in the native runtime.

dtrudg:
_Stacked on #1991_ - content of this PR is in 2nd commit.

When entering a native SIF container, we apply the container-defined environment by sourcing scripts in `/.singularity.d/env` from the relevant action script.

Any environment variable set in these scripts will replace any value set from the OCI runtime config.

To allow `SINGULARITYENV_` / `--env` / `--env-file` to override container-defined env vars, we have to inject them via a script that exports them in `/singularity.d/env`.

We use the script filename `98-singularityenv.sh` as it must:

* Run *after* `95-apps.sh` which sets SCIF App specific variables (which we should be able to override).
* Run *before* `99-base.sh` which mutates `LD_LIBRARY_PATH`.

In addition, we have to consider that:

* In OCI mode, by default we emulate native mode `--compat`. This implies `--no-eval`, where environment variables injected should not be subjected to shell evaluation.
* We have a `--no-compat` flag, that needs to emulate native mode defaults. In this case, environment variables injected should be subjected to shell evaluation.

In general... this is painfully complex, relies on knowledge of where / when various scripts are executed, and may not have the best code structure. However, it works and, the tests enforce the correct behavior. Let's tidy up further later if that's okay? :-)

 - Fixes #1990 

### DIFF
--- a/internal/pkg/runtime/launcher/oci/process_linux.go
+++ b/internal/pkg/runtime/launcher/oci/process_linux.go
@@ -32,7 +32,9 @@ const singularityLibs = "/.singularity.d/libs"
 // Set Singularity> prompt, try bash --norc, fall back to sh.
 var ociShellScript = "export PS1='Singularity> '; test -x /bin/bash && exec /bin/bash --norc || exec /bin/sh"
 
-func (l *Launcher) getProcess(ctx context.Context, imgSpec imgspecv1.Image, bundle string, ep launcher.ExecParams, u specs.User) (*specs.Process, error) {
+// getProcess creates and returns an specs.Process struct defining the execution behavior of the container.
+// The userEnv map returned also holds all user-requested environment variables (i.e. not those from the image).
+func (l *Launcher) getProcess(ctx context.Context, imgSpec imgspecv1.Image, bundle string, ep launcher.ExecParams, u specs.User) (process *specs.Process, userEnv map[string]string, err error) {
 	// Assemble the runtime & user-requested environment, which will be merged
 	// with the image ENV and set in the container at runtime.
 	rtEnv := defaultEnv(ep.Image, bundle)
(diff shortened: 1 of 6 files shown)


internal/pkg/runtime/launcher/oci/process_linux.go
```diff
@@ -32,7 +32,9 @@ const singularityLibs = "/.singularity.d/libs"
 // Set Singularity> prompt, try bash --norc, fall back to sh.
 var ociShellScript = "export PS1='Singularity> '; test -x /bin/bash && exec /bin/bash --norc || exec /bin/sh"
 
-func (l *Launcher) getProcess(ctx context.Context, imgSpec imgspecv1.Image, bundle string, ep launcher.ExecParams, u specs.User) (*specs.Process, error) {
+// getProcess creates and returns an specs.Process struct defining the execution behavior of the container.
+// The userEnv map returned also holds all user-requested environment variables (i.e. not those from the image).
+func (l *Launcher) getProcess(ctx context.Context, imgSpec imgspecv1.Image, bundle string, ep launcher.ExecParams, u specs.User) (process *specs.Process, userEnv map[string]string, err error) {
```
preminger, line 37: We should at some point (not as part of this PR, probably) do something about this overly-large method signature...

preminger (approved):
LGTM. (One non-blocking comment.)

PR #1993, opened 2023-08-04
oci: support --bind of images (bare)

### LINKED ISSUE #1471: Support `--bind` of  images in `--oci` mode (bare)
In native mode, the `--bind` flag has a syntax that allows bind mounting the contents of an image file to a location in the container.

We should support this in `--oci` mode as well. This will be somewhat similar to the `--overlay` case for bare images, in that `squashfuse` or `fuse2fs` will be needed to mount the images to a temporary location. From this location, the images may then be bound into the container. The bind can be added to our runtime config (`config.json`), so that it is actually performed by `runc` / `crun`.

We should be able to support ro and rw binds from bare images here, without issue.

For this PR, implement binding of bare image files (squashfs / extfs).

The syntax is discussed here: https://docs.sylabs.io/guides/latest/user-guide/bind_paths_and_mounts.html#image-mounts

 `-B <image-file>:<dest>:image-src=/`  binds the entire content of the `image-file` onto dest in the container. `-B <image-file>:<dest>:image-src=<source>` binds `source` inside the `image-file` onto `dest` in the container.

The native mode functionality is currently e2e-tested at:

https://github.com/sylabs/singularity/blob/6b76c5958a06fb79b7cd1a86c0a5e88cdf450d53/e2e/actions/actions.go#L1837

preminger:

Add support for bind-mounting bare images in OCI mode (using `-B`/`--bind`, with [identical syntax](https://docs.sylabs.io/guides/latest/user-guide/bind_paths_and_mounts.html#image-mounts) to the corresponding native-mode flag).

 - Fixes #1471

### DIFF
--- a/e2e/actions/oci.go
+++ b/e2e/actions/oci.go
@@ -22,7 +22,9 @@ import (
 	cdispecs "github.com/container-orchestrated-devices/container-device-interface/specs-go"
 	"github.com/sylabs/singularity/v4/e2e/internal/e2e"
 	"github.com/sylabs/singularity/v4/internal/pkg/test/tool/dirs"
+	testExec "github.com/sylabs/singularity/v4/internal/pkg/test/tool/exec"
 	"github.com/sylabs/singularity/v4/internal/pkg/test/tool/require"
+	"github.com/sylabs/singularity/v4/internal/pkg/util/bin"
 	"github.com/sylabs/singularity/v4/internal/pkg/util/fs"
 	"gotest.tools/v3/assert"
 )
@@ -1427,7 +1429,7 @@ func (c actionTests) actionOciOverlay(t *testing.T) {
 
 func haveAllCommands(t *testing.T, cmds []string) bool {
 	for _, c := range cmds {
-		if _, err := exec.LookPath(c); err != nil {
+		if _, err := bin.FindBin(c); err != nil {
 			return false
 		}
 	}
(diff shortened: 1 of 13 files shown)


e2e/actions/oci.go
```diff
@@ -1544,10 +1546,271 @@ func (c actionTests) actionOciOverlayExtfsPerms(t *testing.T) {
...
+	if err := os.Chmod(squashDir, 0o755); err != nil {
+		t.Fatal(err)
+	}
+
+	squashMarkerFile := "squash_marker"
+	if err := fs.Touch(filepath.Join(squashDir, squashMarkerFile)); err != nil {
+		t.Fatal(err)
+	}
+
+	// create the squashfs overlay image
```
dtrudg, line 1599:
```suggestion
	// create the squashfs image
```

e2e/actions/oci.go
```diff
@@ -1544,10 +1546,271 @@ func (c actionTests) actionOciOverlayExtfsPerms(t *testing.T) {
...
+	cmd := testExec.Command("mksquashfs", squashDir, squashfsImage, "-noappend", "-all-root")
+	if res := cmd.Run(t); res.Error != nil {
+		t.Fatalf("Unexpected error while running command.\n%s", res)
+	}
+
+	// create the overlay ext3 image
```
dtrudg, line 1605:
```suggestion
	// create the ext3 image
```

e2e/actions/oci.go
```diff
@@ -1544,10 +1546,271 @@ func (c actionTests) actionOciOverlayExtfsPerms(t *testing.T) {
...
+	if res := cmd.Run(t); res.Error != nil {
+		t.Fatalf("Unexpected error while running command.\n%s", res)
+	}
+
+	cmd = testExec.Command("mkfs.ext3", "-q", "-F", ext3Img)
+	if res := cmd.Run(t); res.Error != nil {
+		t.Fatalf("Unexpected error while running command.\n%s", res)
+	}
+
+	// create new SIF images
```
dtrudg, line 1616: Can leave this code commented until SIF images are needed, but we should slightly adapt the test, and avoid running it for now, as it just wastes CI time.

internal/pkg/util/fs/fuse/fuse_mount_linux.go
```diff
@@ -0,0 +1,182 @@
...
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/sylabs/singularity/v4/internal/pkg/util/bin"
+	"github.com/sylabs/singularity/v4/pkg/image"
+	"github.com/sylabs/singularity/v4/pkg/sylog"
+)
+
+type ImageMount struct {
```
dtrudg, line 19: I think this could become a more generic `fuse.Mount`. While it's an image here so far, that's not going to be generically the case.

internal/pkg/util/fs/fuse/fuse_mount_linux.go
```diff
@@ -0,0 +1,182 @@
...
+	// Readonly represents whether this is a Readonly overlay
+	Readonly bool
+
+	// SourcePath is the path of the image, stripped of any colon-prefixed
+	// options (like ":ro")
+	SourcePath string
+
+	// EnclosingDir is the location of a secure parent-directory in
+	// which to create the actual mountpoint directory
+	EnclosingDir string
```
dtrudg, line 33: This doesn't necessarily make sense now that this is a much more generic fuse mounting routine. The need for a secure parent directory is not universal. It should probably be the callers responsibility.

dtrudg (approved):
This looks good to me. A few thoughts really regarding the fuse mount needing to be more generic now it has moved out of the overlay context to its own package. I will merge and then I'll take any follow up and have others review.

PR #2052, opened 2023-08-15
fuse: look for "fusermount3" first, w/"fusermount" as fallback

### LINKED ISSUE #2047: fuse: must use `fusermount3` or `fusermount` when appropriate
**Version of Singularity**

main

**Describe the bug**

When a host has _only_ fuse 3 installed, then tests and operations will fail:

```
 === RUN   TestAllTypesAtOnce/kerneloverlay
        overlay_set_linux_test.go:89: failed to mount overlay set: use of image "/tmp/TestAllTypesAtOncekerneloverlay1884289746/002/testoverlayitem-1825406996/readonly-extfs.img" as overlay requires fusermount to be installed: exec: "fusermount": executable file not found in $PATH
        testing.go:1225: TempDir RemoveAll cleanup: unlinkat /tmp/TestAllTypesAtOncekerneloverlay1884289746/001/testoverlayitem-2352311219/upper: read-only file system
    === RUN   TestAllTypesAtOnce/fuseoverlayfs
```

We need to use `fusermount3` not `fusermount` when `fusermount` is not available.

preminger:

Change how we search for the external tool `fusermount` so that we first look for `fusermount3` and then search for `fusermount` as a fallback.

Also updates the require.Command() function in internal/pkg/test/tool/require/require.go so that it first tries to find the command using the bin.FindBin() function, and only if that fails, tries exec.LookPath(). This allows us to implicitly incorporate the fallback logic of bin.FindBin() when gating tests using require.Command() (so, for example, after this PR, `require.Command("fusermount")` will allow the test to run even if only `fusermount3` is installed, and vice versa).

 - Fixes #2047

### DIFF
--- a/internal/pkg/test/tool/require/require.go
+++ b/internal/pkg/test/tool/require/require.go
@@ -22,6 +22,7 @@ import (
 	"github.com/opencontainers/runc/libcontainer/cgroups"
 	"github.com/sylabs/singularity/v4/internal/pkg/buildcfg"
 	"github.com/sylabs/singularity/v4/internal/pkg/security/seccomp"
+	"github.com/sylabs/singularity/v4/internal/pkg/util/bin"
 	"github.com/sylabs/singularity/v4/internal/pkg/util/rpm"
 	"github.com/sylabs/singularity/v4/pkg/network"
 	"github.com/sylabs/singularity/v4/pkg/util/fs/proc"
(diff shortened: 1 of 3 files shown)


internal/pkg/test/tool/require/require.go
```diff
@@ -258,7 +259,7 @@ func Filesystem(t *testing.T, fs string) {
 // in one the path defined in the PATH environment variable,
 // if not found the current test is skipped with a message.
 func Command(t *testing.T, command string) {
-	_, err := exec.LookPath(command)
+	_, err := bin.FindBin(command)
```
dtrudg, line 262: Might need to check that this doesn't skip more tests.

We are quite likely to be using `require.Command` for something in a test that *isn't* expected to be used by Singularity.

preminger: Thanks, good point - changing it so that it tries bin.FindBin() and if that fails, tries exec.LookPath(). Only if both fail, will t.Skipf() actually be called.

dtrudg (changes requested):
LGTM - just 2 nits.

dtrudg (approved):
LGTM

PR #2065, opened 2023-08-16
lint: enable forcetypeassert + corresponding fixes

preminger:

Enable the `forcetypeassert` linter in golangci-lint, and perform fixes / add `nolint:` directives, as appropriate, in code.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -26,6 +26,7 @@ linters:
     - dupl
     - dupword
     - exportloopref
+    - forcetypeassert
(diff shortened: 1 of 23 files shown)


dtrudg (changes requested):
I would suggest adding type assertion checks on the `pkg/flag` stuff.

It's a package we are exporting and it would be good to have a better guard against a flag of the wrong type being passed into the functions.

dtrudg (approved):

PR #2066, opened 2023-08-16
lint: enable gosec + corresponding fixes

preminger:

Enable the `gosec` linter in golangci-lint, and perform fixes / add `#nosec` directives, as appropriate, in code.

Also adds three specific `gosec` rules to golangci-lint "exclude" list due to large number of false positives.


dtrudg (approved):
Looks good to me generally.

If you could add comments in the .golang.ci.yml mentioning what the disabled rules are, and why they are disabled, that would be nice for people coming across them.

PR #2101, opened 2023-08-18
fix: make: change cond for setting GO_RACE to handle i386 & i686

preminger:

The `-race` to `go test` is not supported on `i686` (on platforms like Raspberry Pi OS).

This PR changes the test that suppresses `-race` in mlocal/frags/go_common_opts.mk so it includes not only `ppc64le` but also, disjunctively, `i386' and `i686` as well.


dtrudg (approved):
LGTM. Thanks!

PR #2105, opened 2023-08-22
add better handling of cmd output in squashfs code

preminger:

This PR improves the way the output of external commands is handled inside internal/pkg/util/fs/squashfs/squashfs.go - in particular, if an external command exits with an error, the error propagated to the caller will include the stdout and stderr output the command.


tri-adam (approved):
LGTM

PR #2130, opened 2023-08-29
oci: fallback to temp dir if OCI-SIF unsupported

### LINKED ISSUE #2095: oci: fallback to temporary bundle if no sqfstar / tar2sqfs?
**Version of Singularity**

main

**Describe the bug**

In Singularity 3.11, the experimental `--oci` mode uses a temporary container bundle under `/tmp` rather than creating an OCI-SIF for any OCI container to be executed.

OCI-SIF has many advantages, but requires `sqfstar` / `tar2sqfs`. If these are not present, OCI-mode cannot be used. This could be considered a regression from 3.11

We do still have all of the code present to use temporary bundle under `pkg/ocibundle/native`.

Should we add a path to fall back to this, in the case that  `sqfstar` / `tar2sqfs` are not available... to avoid a regression from 3.11?

preminger:

In OCI-mode, if using OCI-SIF fails (e.g. `squashfuse`/`squashfuse_ll` missing or don't work, `sqfstar`/`tar2sqfs` missing, etc.), fall back to a filesystem-based strategy: unpack the OCI container into a temporary directory, and run it from there.

Add an `tmp sandbox` configuration option to `singularity.conf` (default: `yes`) to control whether unpacking containers to temporary sandbox dirs is allowed, as well as a `--no-tmp-sandbox` flag to disable this behavior via the command-line. NOTE: This configuration option & flag also control whether unpacking the container to a temporary sandbox dir is allowed in the native user-namespace flow.

 - Fixes #2095

### DIFF
--- a/cmd/internal/cli/actions.go
+++ b/cmd/internal/cli/actions.go
@@ -8,6 +8,7 @@ package cli
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -25,6 +26,7 @@ import (
 	"github.com/sylabs/singularity/v4/internal/pkg/runtime/launcher/native"
 	ocilauncher "github.com/sylabs/singularity/v4/internal/pkg/runtime/launcher/oci"
 	"github.com/sylabs/singularity/v4/internal/pkg/util/uri"
+	"github.com/sylabs/singularity/v4/pkg/ocibundle/ocisif"
 	"github.com/sylabs/singularity/v4/pkg/syfs"
 	"github.com/sylabs/singularity/v4/pkg/sylog"
 	useragent "github.com/sylabs/singularity/v4/pkg/util/user-agent"
@@ -45,6 +47,12 @@ func getCacheHandle(cfg cache.Config) *cache.Handle {
 
 	return h
 }
+
+type contextKey int
+
+const (
+	keyOrigImageURI contextKey = iota
+)
 
 // actionPreRun will:
 //   - do the proper path unsetting;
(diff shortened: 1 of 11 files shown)


cmd/internal/cli/actions.go
```diff
@@ -46,6 +48,12 @@ func getCacheHandle(cfg cache.Config) *cache.Handle {
 	return h
 }
 
+type contextKey int
+
+const (
+	keyOrigImageURI contextKey = iota
```
dtrudg, line 54: Not clear why you are using an int key here? A string const key would be more usual I think.

dtrudg (changes requested):
Functionally this LGTM.

However, the addition of an option and flag to prevent fall-back to directory extraction makes me think of the user-namespace flow in native mode which does a somewhat similar extraction to run from a temporary directory:

```
$ singularity run -u docker://alpine
Singularity> 
```

I think it would be better if a generic option/flag suppressed both the OCI flow here, and the user namespace native mode flow. That way, if we add squashfuse etc. mounting to unprivileged native mode (which is on the cards for 4.1), we have a single common option/flag that applies for oci and native modes.

dtrudg (approved):
LGTM. The warning message is a bit long :grimacing: ... but that's due to the depth and wrapping here.

Let's merge, and if we get to tidying that up somewhat it'd be a bonus.

```
WARNING: OCI-SIF functionality could not be used, falling back to unpacking OCI bundle in temporary sandbox dir (original error msg: while mounting squashfs layer: failed to mount: fork/exec /usr/local/libexec/singularity/bin/squashfuse_ll: no such file or directory (cmdline: "/usr/local/libexec/singularity/bin/squashfuse_ll -o ro,offset=32176,uid=0,gid=0 /home/dtrudg-sylabs/.singularity/cache/oci-sif/sha256:c5c5fda71656f28e49ac9c5416b3643eaa6a108a8093151d6d1afc9463be8e33 /usr/local/var/singularity/mnt/session/oci-bundle2003726521/rootfs"; output: ""))
```

PR #2159, opened 2023-09-05
doc: Update LICENSE_DEPENDENCIES.md generation

dtrudg:

Upgrade go-licenses to the latest 1.6.0 which now handles the UNLICENSE properly, etc.

Use a template approach for the markdown generation, rather than converting from CSV ourselves.

The template includes full license text, per-dependency, rather than a link to the URL for the dependency's license. This is useful as it:

* Ensures we are distributing the required license.
* Avoids versioned links that change at each dependency update, and require manual resolution for dependabot updates.


preminger (approved):
LGTM

PR #2183, opened 2023-09-13
detect and advise re:OCI-SIF images on older versions (release-3.11)

### LINKED ISSUE #2097: release-3.11: detect and advise r.e. OCI-SIF images?
It may be useful to update the release-3.11 branch so that it:

* Is able to detect OCI-SIF images (backport of some pkg/image code from 4.0.0)
* Produces a useful error message when presented with an OCI-SIF image, stating that these require Singularity 4.0.0.

If we have this in a final 3.11.5 that gets into EPEL etc, then it should reduce the inquiries we get from people trying to use OCI-SIF on an older install.

preminger:

Detect OCI-SIF images and advise the user that they are only supported on Singularity versions 4.0 and above.

This PR cherry-picks https://github.com/sylabs/singularity/pull/2183/commits/084fed552decb9f92ddaea10892353896aed8aa9 ("pkg/image: Add minimal oci-sif type") from 4.0, and then adds code specific to the `release-3.11` branch in a second commit.

In that second commit, the changes in internal/pkg/runtime/launcher/native/launcher_linux.go are not strictly speaking necessary, as the changes to pkg/ocibundle/native/bundle_linux.go would be sufficient to catch the relevant condition as it stands in the `release-3.11` branch right now. But the changes to internal/pkg/runtime/launcher/native/launcher_linux.go are there to guard against potential future scenarios in which we add new code flows to `release-3.11` that may reach those points in the code without going through the relevant part of pkg/ocibundle/native/bundle_linux.go first.

 - Fixes #2097

### DIFF
--- a/internal/pkg/runtime/launcher/native/launcher_linux.go
+++ b/internal/pkg/runtime/launcher/native/launcher_linux.go
@@ -486,7 +486,7 @@ func (l *Launcher) checkImage() error {
 	}
 
 	if img.Type == image.OCISIF {
-		return fmt.Errorf("native runtime does not support OCI-SIF images, use --oci mode")
+		return fmt.Errorf("encountered OCI-SIF image, which is only supported on singularity versions 4.0 and above")
 	}
 
 	if err := l.checkEncryptionKey(img); err != nil {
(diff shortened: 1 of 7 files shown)


dtrudg (approved):
LGTM

PR #2188, opened 2023-09-14
determine OCI-SIF vs. native SIF using image contents, not URI

### LINKED ISSUE #2184: remove brute-force oci-sif vs. native sif determination...?
It seems like the code we now have in `pkg/image` satisfies the precondition in this TODO:

https://github.com/sylabs/singularity/blob/76ccdddf6323fd11c9444c8251d3474570c93ced/cmd/internal/cli/actions.go#L207C8-L207C8

```golang

	// TODO - drop once we have implemented prefix-less oci-sif vs native sif detection.
	if isOCI {
		image = "oci-sif:" + image
	}

```

Is that correct? And if so, is it time to address this?

preminger:

The code in `pkg/image` is now able to detect OCI-SIF images (and distinguish them from native SIFs) based on image content. This PR replaces reliance on the URI transport field being `oci-sif:` with detection of OCI-SIF images based on the image content.

 - Fixes #2184


cmd/internal/cli/actions.go
```diff
@@ -200,15 +201,6 @@ func replaceURIWithImage(ctx context.Context, cmd *cobra.Command, args []string)
 		return origImageURI
 	}
 
-	if err != nil {
```
dtrudg, line 203: Don't think this should be dropped?

It is handling where one of the `case`s returns an error which is not a OCI-SIF mount issue.

internal/pkg/util/image/ocisif.go
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2023, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package image
```
dtrudg, line 6: I don't see a good reason not to have this as a public convenience function in `pkg/image`?

I don't think it is at risk of an API change.

dtrudg (changes requested):

cmd/internal/cli/actions.go
preminger: Yeah, not sure how that ended up being part of what I removed...

pkg/image/ocisif.go
```diff
@@ -60,3 +61,13 @@ func (f *ociSifFormat) openMode(bool) int {
 func (f *ociSifFormat) lock(*Image) error {
 	return nil
 }
+
+func IsOCISIF(filename string) (bool, error) {
```
dtrudg, line 65: As this is an importable pkg, there should be a docstring on the function.

dtrudg (approved):
LGTM - but please add a doc string on the function in `pkg`.

PR #2199, opened 2023-09-15
WIP: de-duplicate fuse mounting/unmounting code

### LINKED ISSUE #2025: chore: de-duplicate fuse mounting/unmounting code
**Type of issue**

technical debt

**Description of issue**

Following work on the OCI mode, we now have `internal/pkg/util/fs/fuse` as a place for FUSE mount/unmount related code.

We are performing FUSE mounts in various other places, using different code.

We should expand `internal/pkg/util/fs/fuse` so that it can be used to replace other FUSE mount/unmount code, to reduce the number of unique places that may need to be examined given any FUSE mount bugs / maintenance.

preminger:

Transition as much FUSE-related code as is currently feasible to using internal/pkg/util/fs/fuse package.

 - Fixes #2025


internal/pkg/runtime/engine/singularity/cleanup_host_linux.go
```diff
@@ -20,18 +19,7 @@ import (
...
-		}
-
-		err = sifuser.Unmount(ctx, e.EngineConfig.GetImage(),
-			sifuser.OptUnmountStdout(os.Stdout),
-			sifuser.OptUnmountStderr(os.Stderr),
-			sifuser.OptUnmountFusermountPath(fusermountPath))
-		if err != nil {
-			return fmt.Errorf("while unmounting fuse directory: %s: %w", e.EngineConfig.GetImage(), err)
-		}
+		squashfs.FUSEUnmount(ctx, e.EngineConfig.GetImage())
```
dtrudg, line 22: I think we definitely need to handle an error here. If we are unable to unmount in this flow, we will accumulate mounts on the host. The failure should be visible so that the underlying issue has the chance of reaching an admin and being fixed.

internal/pkg/runtime/launcher/native/launcher_linux.go
```diff
@@ -1156,18 +1157,20 @@ func squashfuseMount(ctx context.Context, img *imgutil.Image, imageDir string) (
...
 
-	squashfusePath, err := bin.FindBin("squashfuse")
+	f, err := sif.LoadContainerFromPath(img.Path, sif.OptLoadWithFlag(os.O_RDONLY))
 	if err != nil {
-		return fmt.Errorf("squashfuse is required: %w", err)
+		return fmt.Errorf("failed to load image: %w", err)
 	}
-	if _, err := bin.FindBin("fusermount"); err != nil {
-		return fmt.Errorf("fusermount is required: %w", err)
+	defer func() { _ = f.UnloadContainer() }()
```
dtrudg, line 1164: Not sure that we want to do this here. `f` will be an open file handle to the SIF outside of any namespaces etc. This clearly shows that it is 'in use' by singularity, keeps any automounter happy etc.

Arguably we would actually benefit from performing the mount against the file handle that's open at this point... to remove the chance of the SIF being moved between here and the FUSE mount call.

internal/pkg/util/fs/fuse/fuse_mount_linux.go
```diff
@@ -43,11 +44,15 @@ type ImageMount struct {
 
 	// AllowOther is set to true to mount the image with the "allow_other" option.
 	AllowOther bool
+
+	// ExtraMountOpts are options to be passed to the mount command (in the "-o"
+	// argument) beyond the ones autogenerated from other ImageMount fields.
+	ExtraMountOpts []string
```
dtrudg, line 50: Either `Opts` or `ExtraOpts`. Mount is redundant given this is `ImageMount.xxxx`

internal/pkg/util/fs/fuse/fuse_mount_linux.go
```diff
@@ -142,6 +147,7 @@ func (i *ImageMount) generateCmdArgs() ([]string, error) {
 	if i.AllowOther {
 		opts = append(opts, "allow_other")
 	}
+	opts = append(opts, i.ExtraMountOpts...)
```
dtrudg, line 150: We should check that `ExtraMountOpts` does not contain any options that are handled by other fields.

E.g. `ro/rw` `nosetuid` etc. should not be permitted as an extra option. We need to be able to rely on the fact that e.g. inspecting `.ReadOnly` or `.AllowSetuid` gives us the true state of the mount.

This is potentially critical for security and should be tested.

dtrudg (changes requested):
Taking shape nicely. I've left some comments. My chief concern is that extra options must not be allowed to potentially override things that are set from specific attributes of ImageMount.

internal/pkg/util/fs/fuse/fuse_mount_linux.go
```diff
@@ -124,33 +131,95 @@ func (i *ImageMount) generateCmdArgs() ([]string, error) {
...
+	args = append(args, i.SourcePath)
+	args = append(args, i.mountpoint)
+
+	return args, nil
+}
+
+func (i ImageMount) generateMountOpts() ([]string, error) {
 	// TODO: Think through what makes sense for file ownership in FUSE-mounted
 	// images, vis a vis id-mappings and user-namespaces.
 	opts := []string{"uid=0", "gid=0"}
```
dtrudg, line 152: I think this will result in 2 `uid=n,gid=n` strings in the squashfs flow which sets the user's uid and gid? It appears to work, but maybe uid, gid should become attributes. They'd naturally default to 0 if stored as int / uint.

dtrudg (commented):
This is LGTM - but I think I missed something r.e. uid / gid handling which should probably at this point mean that they are handled as attributes?

internal/pkg/util/fs/fuse/fuse_mount_linux.go
```diff
@@ -24,6 +24,12 @@ type ImageMount struct {
 	// values in pkg/image)
 	Type int
 
+	// UID is the value to pass to the uid option when mouting
```
dtrudg, line 27: `mounting`

internal/pkg/util/fs/fuse/fuse_mount_linux.go
```diff
@@ -24,6 +24,12 @@ type ImageMount struct {
...
+	UID int
+
+	// GID is the value to pass to the gid option when mouting
```
dtrudg, line 30: `mounting`

dtrudg (approved):
LGTM - noted some typos in comments

PR #2213, opened 2023-09-20
fix: Don't mount /var/tmp pointing to /tmp in container

dtrudg:
## Description of the Pull Request (PR):

It is not uncommon for `/var/tmp` in a container to be a symlink onto `/tmp`. In this case, singularity will currently:

* Mount host `/tmp` onto the container `/tmp`
* Mount host `/var/tmp` onto the container `/tmp` through the symlink.

Check whether `/var/tmp` resolves to the same place as `/tmp` in the container, and do not mount it if this is the case.

We have to move `addTmpMount` to run under a `RunBeforeTag` call because we need to have the container rootfs in place to to inspect for symlinks.

### This fixes or addresses the following GitHub issues:

 - Fixes #1950
 - Fixes #1331 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/main/CONTRIBUTORS.md)

PR #2221, opened 2023-09-21
oci: fix: Don't mount /var/tmp pointing to /tmp in container

preminger:

This PR is the OCI-mode counterpart of native-mode PR #2213 

It is not uncommon for `/var/tmp` in a container to be a symlink onto `/tmp`. In this case, singularity will currently:

* Mount host `/tmp` onto the container `/tmp`
* Mount host `/var/tmp` onto the container `/tmp` through the symlink.

At the end of finalizeSpec() (internal/pkg/runtime/launcher/oci/launcher_linux.go:406), check if in-container `/tmp` and `/var/tmp` resolve to the same location. If so, remove any default mounts (but _not_ user-generated ones) whose destination is `/var/tmp` from the OCI spec we're constructing.

Also adds e2e tests to ensure parity of behavior across native- and OCI-modes with user-requested bind mounts of /var/tmp, with `(--contain) --workdir`, and with `(--compat)`. (Parentheses in the preceding text indicate independently-necessary differences between native- and OCI-mode command line options to ensure a common baseline for the tests.)


internal/pkg/runtime/launcher/oci/launcher_linux.go
```diff
@@ -397,9 +403,26 @@ func (l *Launcher) finalizeSpec(ctx context.Context, b ocibundle.Bundle, spec *s
...
+	varTmpResolved := fs.EvalRelative(vartmpPath, spec.Root.Path)
+	sylog.Debugf("Container /tmp resolves to %q", tmpResolved)
+	sylog.Debugf("Container /var/tmp resolves to %q", varTmpResolved)
+
+	if varTmpResolved != tmpResolved {
+		return
+	}
+
+	spec.Mounts = lo.Filter(spec.Mounts, func(m specs.Mount, _ int) bool {
+		return (m.Destination != vartmpPath)
```
dtrudg, line 422: Sorry - thinking about this implementation is giving me pause for thought again. r.e. explicit binds of both `-B /tmp` and `-B /var/tmp` where there is a symlink.

This filtering here is different than the native mode, because the native mode exclusion of `/var/tmp` is only for the automatic mount, not a user requested bind.

I need to go back and look at Docker / Podman behaviours etc. and get it straight in my head. It might be correct as-is.... apologies for delaying this. My fault!

dtrudg: I've thought through this now, and looked at Docker in the gist here: https://gist.github.com/dtrudg/4dda5e5e2d3c01fa00ae1f040a5ebeb6

We can't do a simple filter by destination for anything mounted onto `/var/tmp`, as this removes both the default and any explicit user-requested `/var/tmp` binds... which must be handled differently:

The default mount onto `/var/tmp` must be skipped if `/var/tmp` is a symlink to `/tmp`. This is mostly important in OCI-mode in conjunction with `--workdir`... to ensure `workdir/tmp` and not `workdir/var_tmp` is present at `/tmp` in the container.

If a user requests `-B /var/tmp` then it must be honoured, and never filtered out, even if it is going to mask something else on `/tmp` via the symlink.

dtrudg (changes requested):
Please see comment above.

internal/pkg/runtime/launcher/oci/launcher_linux.go
```diff
@@ -397,9 +407,40 @@ func (l *Launcher) finalizeSpec(ctx context.Context, b ocibundle.Bundle, spec *s
...
+	})
+
+	// Remove default mounts whose destination is /var/tmp from the OCI spec
+	// we're generating.
+	for _, defMount := range vartmpDefaultMounts {
+		spec.Mounts = lo.Filter(spec.Mounts, func(m specs.Mount, _ int) bool {
+			// The first disjunct here is superfluous, logically speaking; it's
+			// here to ensure comparison is as fast as possible. It only makes
+			// sense to bother with DeepEqual() if we know the path matches.
+			return ((m.Destination != vartmpPath) || !reflect.DeepEqual(m, defMount))
```
dtrudg, line 439: I'm not sure whether `reflect.DeepEqual` is the best thing to filter with here.

In theory, could someone specify a `-B` bind of a workdir var_tmp that would end up with the same struct content as a default bind from a workdir var_tmp?

Also... if we extend our Docker compatible `--mount` syntax to support things other than binds, then it'd be possible for someone to request a tmpfs mount via `--mount`.... which perhaps would be indistinguishable from the 

Can we filter based on the address of the default mount spec.Mount structs... rather than their content, to avoid any issue?

preminger: Yeah, good point!

dtrudg (approved):
This LGTM. Additional tests very welcome (!)

It could probably be optimized down to just storing the default mount `/var/tmp` index only, and then doing the removal as a one-liner:

```
spec.Mounts = append(spec.Mounts[:index], spec.Mounts[index+1:]...)
```

But am happy for it to stay as is.

preminger:
> It could probably be optimized down to just storing the default mount `/var/tmp` index only [...]

Yeah, so, as I think I mentioned on Slack (?), this was intentionally done this way —

It doesn't seem inconceivable (to me, anyway) that in the future there would be circumstances where, depending on the combination of command-line flags involved, more than one default-generated mount targeting `/var/tmp` would be created in some situations. (Obviously, the last one enacted would be the one that's effective; unless overlays were somehow involved. But that doesn't mean there necessarily won't be circumstances where `spec.Mounts` nevertheless contains more than one such entry.)

The way this is currently written, we don't have to worry about that, moving forward. We now have a way of quickly fetching everything that was added by addTmpMount() and only that. I think that's good to have.

PR #2222, opened 2023-09-22
oci: fix: Don't mount /var/tmp pointing to /tmp in container (release-4.0)

preminger:

Pick #2221 

This PR is the OCI-mode counterpart of native-mode PR #2213 

It is not uncommon for `/var/tmp` in a container to be a symlink onto `/tmp`. In this case, singularity will currently:

* Mount host `/tmp` onto the container `/tmp`
* Mount host `/var/tmp` onto the container `/tmp` through the symlink.

At the end of finalizeSpec() (internal/pkg/runtime/launcher/oci/launcher_linux.go:406), check if in-container `/tmp` and `/var/tmp` resolve to the same location. If so, remove any default mounts (but _not_ user-generated ones) whose destination is `/var/tmp` from the OCI spec we're constructing.

Also adds e2e tests to ensure parity of behavior across native- and OCI-modes with user-requested bind mounts of /var/tmp, with `(--contain) --workdir`, and with `(--compat)`. (Parentheses in the preceding text indicate independently-necessary differences between native- and OCI-mode command line options to ensure a common baseline for the tests.)

### DIFF
--- a/e2e/actions/regressions.go
+++ b/e2e/actions/regressions.go
@@ -705,24 +705,54 @@ func (c actionTests) issue1950(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	varTmpHostResolved, err := filepath.EvalSymlinks("/var/tmp")
+	vartmpHostResolved, err := filepath.EvalSymlinks("/var/tmp")
 	if err != nil {
 		t.Error(err)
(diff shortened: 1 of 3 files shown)


dtrudg (approved):
👍

PR #2223, opened 2023-09-22
e2e: fix profile in issue1950() regression "native binds" subtest

preminger:

Corrects the `profile` field of the "native binds" subtest of the issue1950() e2e regression test.


dtrudg (approved):
Good catch - sorry I missed that!

Issue #2022, opened 2023-08-11
deps: remove containers/image direct dependency


dtrudg:
**Type of issue**

technical debt

**Description of issue**

Since Singularity 3.0, we've used the `containers/image` package to handle the retrieval of container images from OCI sources.

We also use types from `containers/image` to pass around authentication information, platform preferences, etc.

`containers/image` generally works well but...

* It has a very long dependency list, causing singularity to have more indirect dependencies than would otherwise be the case.
* It handles progress bars and logging for OCI operations directly. This means it doesn't fit very well with other CLI output /logging in Singularity.
* Because it is targeted at the needs of the podman ecosystem, it includes some 'magic' handling of system/user configuration files that make sense for podman. The effects of these can be somewhat difficult to identify, and can be confusing to a user who does not expect podman ecosystem configuration to affect the bahaviour of Singularity.

With the work to support OCI-SIF, we have now added quite a lot of code using `google/go-containerregistry`. This is a much lighter-weight module than `containers/image`, with less magic. It should be relatively straightforward to replace our `containers/image` code with code that uses `google/co-containerregistry.` This would bring us back to the situation of using one module for OCI interactions, and reduce our dependency tree.

PR #2225, opened 2023-09-25
add --authfile to OCI login/logout, pull, push, and action-cmds

### LINKED ISSUE #2098: --authfile flag for OCI credentials

Sometimes it is necessary to use time-limited, or specific-purpose, credentials to push/pull to/from OCI registries.

SingularityCE does support using environment variables, but sometimes a file storing credentials is more convenient.

Other runtimes have an `--authfile` flag that allows a specific file to be used as the store of OCI credential, rather than the default. We should provide this functionality in SingularityCE.

The `--authfile` flag would need to be available on the `singularity registry` command, and any other commands that consume credentials.

Some thought needs to be given to the way in which we currently store OCI login information split between `remote.yml` and the credentials file (which is in docker config format). How should this split be reconciled with an `--authfile` flag?

preminger:

Adds a `--authfile <path>` flag to the `registry login` and `registry logout` commands, which causes the OCI credentials to be written to / removed from a custom file located at `<path>` instead of the default location (`$HOME/.singularity/docker-config.json`).

Adds an identically named `--authfile <path>`  flag to the `pull`, `push`, `run`, `exec`, `shell`, and `instance start` commands, to read OCI registry credentials from a custom file written with `registry login --authfile <path>` instead of from the default location.

 - Fixes #2098


cmd/internal/cli/singularity.go
```diff
@@ -319,6 +322,15 @@ var commonPlatformFlag = cmdline.Flag{
 	EnvKeys:      []string{"PLATFORM"},
 }
 
+// --authfile
+var commonAuthFileFlag = cmdline.Flag{
+	ID:           "commonAuthFileFlag",
+	Value:        &ociAuthFile,
+	DefaultValue: "",
```
dtrudg, line 329: It's not clear to me why we don't default to the `syfs.DockerConf` value here?

internal/pkg/client/ocisif/ocisif.go
```diff
@@ -14,6 +14,8 @@ import (
 	"strings"
 	"time"
 
+	ociref "github.com/containers/image/v5/docker/reference"
+	ociconfig "github.com/containers/image/v5/pkg/docker/config"
```
dtrudg, line 18: We definitely want to avoid adding `containers/image` dependencies here. if possible.

This package is `containers/image` own implementation of the docker config file that we are already using `docker/cli` for elsewhere.

internal/pkg/client/ocisif/ocisif.go
```diff
@@ -281,5 +283,18 @@ func PushOCISIF(_ context.Context, sourceFile, destRef string, ociAuth *ocitypes
 		return fmt.Errorf("while obtaining image: %w", err)
 	}
 
+	if ociAuthFile != "" {
```
dtrudg, line 286: Instead of this block, we need to modify the AuthOptn function that is used below, so it can pull the credential out of a file at a custom path.

That code is in: https://github.com/preminger/singularity/blob/bfb7a87dd808dde5798da88cfe0061dcc209221c/internal/pkg/client/ocisif/auth.go#L102

internal/pkg/remote/credential/login_handler.go
```diff
@@ -87,6 +90,14 @@ func (h *ociHandler) login(u *url.URL, username, password string, insecure bool)
 		return nil, err
 	}
 
+	if ociAuthFile != "" {
```
dtrudg, line 93: There shouldn't be any need for this additional block that's bringing in `containers/image`. A custom auth file should be handled in the same flow we are using for `~/.singularity/docker-config.json`

You should be able to handle a different file through changing `ociConfig` and `ociConfigNew` below.

dtrudg (changes requested):
Let's discuss this one today.

There should be no reason to add `containers/image` code to write the custom credential file. We are already using `docker/cli` to write the default credential file,  which is being consumed by both `containers/image` and `google/go-containerregistry` code.

cmd/internal/cli/singularity.go
preminger: The reasoning is as follows: when the user did not specify an explicit value & the `~/.singularity/docker-config.json` file cannot be found, it makes sense to fall back to anonymous authentication (with an appropriate message to `sylog`). But if a user specified a path to an authfile and that authfile cannot be found, I _don't_ think the same behavior makes sense; it makes more sense to error out in this circumstance, because the user clearly wanted specifically the authentication located where (they thought) the file is.

This means we have to detect whether an `--authfile` argument was or wasn't passed. If we set the default to `syfs.DockerConf()`, we would have no way of distinguishing passing "~/.singularity/docker-config.json" as an argument (however much of an edge case that may be) from passing no argument at all.

internal/pkg/remote/credential/login_handler.go
preminger: Underlying cause is #2226; marking as resolved.

internal/pkg/client/ocisif/ocisif.go
preminger: Underlying cause is #2226; marking as resolved.

cmd/internal/cli/singularity.go
dtrudg: I still believe there is value in using the default value on the flag, as it would then allow someone to explicitly set `--authfile ''` to avoid authentication at all. That's a lot less disruptive than doing `registry logout`, and less hassle than manually creating a null authfile to use.

This is, I believe, a much more likely case that someone specifying singularity's managed `~/.singularity/docker-config.json' file themselves.

internal/pkg/util/ociauth/ociauth.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2023, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package ociauth
```
dtrudg, line 6: I'm not sure there is a good reason not to keep this with other OCI related code in `pkg/remote/credential`?

dtrudg (dismissed):
~~Please see https://github.com/sylabs/singularity/pull/2225#discussion_r1338324792~~

~~Given that the customer who requested this is also looking for `--authfile` to work like podman, I do think doing it through a flag default that can be overridden to "" is relatively important.~~

~~https://github.com/containers/podman/blob/35824b647cc87e11ad6712ecaabafcacc3b66e96/cmd/podman/common/create.go#L54~~

Edit - scratch this thought.... in podman it's more complicated with handling buried in `containers/common` and `containers/image` such that a null `--authfile` doesn't actually work as it might first appear.

Their `GetDefaultAuthFile` doesn't actually return the usual default auth file. Heh.

internal/pkg/util/ociauth/ociauth.go
preminger: Just a general hesitancy to put things in toplevel `pkg/` without having a good long think about it first...

preminger: Sorry, now realizing that you meant `*internal*/pkg/remote/credential` - in which case, yes, there's a reason: the stuff in ociauth.go also gets used in internal/pkg/client/ocisif/ocisif.go, so it made sense (to me at least) to have it in a place that's related to both consumers (internal/pkg/client/ocisif/ocisif.go and internal/pkg/remote/credential/login_handler.go) in a symmetrical way.

dtrudg (changes requested):
Let's discuss this one today. There are two main issues:

1) Moving the credential storage code into `pkg/internal/client/ocisif`. This package is intended to be for code that deals only with push/pull of OCI-SIF images. There's some auth code in there because it uses gccr while most other stuff is containers/image. However, the general direction should be that the auth code comes out of this package over time... into a package specific to auth, as we move away from containers/image.

     Another disadvantage with the code being moved is that this feature is highly likely to be a backport request to 3.11 for a PRO customer. There is no `pkg/internal/client/ocisif` there, which makes a backport more awkward to perform and maintain.

2) The auth code that was already existent in `pkg/internal/client/ocisif` was written to mirror that in gccr itself, modified to allow us to use our custom auth file.

     https://github.com/google/go-containerregistry/blob/main/pkg/authn/keychain.go

    This allows us to easily track / refer to any changes in gccr's own implementation etc.
    
    In this PR the code is now dramatically different, which hampers maintainability and is at odds with the planned #2022 

In addition, I think the e2e approach has been over-complicated w.r.t. the attempts at tracking / locking auth etc. Using a profile's `.String()` is not robust for scoping login/logout, given that some profiles share the same home directory and default auth file. However it's not an issue here because you are putting the action and registry tests in `np(..)` so they run sequentially. Any code to facilitate parallelism needs to be robust, and used. If it's not really exercised due to np test it'd be better to just remove the complexity and annotate login/logout functions with a function comment indicating they are not concurrency safe.

internal/pkg/util/ociauth/ociauth.go
dtrudg: Sorry about the missing `internal`.

Code that's called from two places doesn't always have to be moved out to a third place for caller symmetry. The package `internal/pkg/remote/credential` is specifically about handling/storing OCI and other credentials, so is an appropriate place.

In Go code packages are often quite large, and things are internally split through multiple source files within the package, rather than many smaller packages being created.

I would agree about splitting, however, if the code is buried in the UI layer. I.E. if this code had been buried in the `cmd` or `app` portions then I'd definitely have agreed about moving it... as we want a logical separation of interface/CLI and lower level implementation stuff.

dtrudg (approved):
LGTM. Thanks particularly for the e2e work. Good to have coverage where there were large gaps previously.

PR #2230, opened 2023-09-29
e2e: move aside existing remote.yaml files in remoteUseExclusive()

preminger:

The e2e test "TestE2E/SEQ/REMOTE/use_exclusive" (`remoteUseExclusive()`, e2e/remote/remote.go:563) depends on there not being a remote.yaml in the root user's homedir prior to the test being run.

This PR adds code that moves existing remote.yaml in users' homedir aside for the duration of this test, and moves them back into place as part of the test's cleanup.

### DIFF
--- a/e2e/remote/remote.go
+++ b/e2e/remote/remote.go
@@ -567,33 +567,28 @@ func (c ctx) remoteUseExclusive(t *testing.T) {
 	)
 
 	// Move the user's and root's remote.yaml files aside for the purposes of this test
-	asideFiles := make(map[uint32]string)
+	origRelPath := filepath.Join(".singularity", "remote.yaml")
+	asideRelPath := fmt.Sprintf("%s.aside-remoteUseExclusive", origRelPath)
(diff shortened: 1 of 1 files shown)


dtrudg (changes requested):
This works, but it's really over-engineered - to the extent it's harder than it needs to be to verify it's doing what I think, from quickly scanning the code.

If the moveAside and restore functions were separate named functions that could be called from other tests then the overall structure would make sense to me.

As it is... with them completely scoped to this particular test...

* A rename with a `.orig` suffix, and back again, would probably be sufficient. It would also be a true rename, rather than a copy+delete (as the home and /tmp

* The move doesn't need to be in a separate func() and the restore could just be inside the t.Cleanup(func(

Please could you either...

* Make the move/restore functions properly generic e2e utility functions -or-
* Drop the complexity to a simple rename to a known suffix, and back again in a cleanup.

dtrudg (approved):

PR #2238, opened 2023-10-05
build: add support for --authfile flag

preminger:

Add support for the `--authfile` flag for (non-remote) `build` command.

### DIFF
--- a/cmd/internal/cli/build_linux.go
+++ b/cmd/internal/cli/build_linux.go
@@ -147,6 +147,10 @@ func runBuild(cmd *cobra.Command, args []string) {
 		os.Setenv("SINGULARITY_WRITABLE_TMPFS", "1")
 	}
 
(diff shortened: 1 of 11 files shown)


cmd/internal/cli/build_linux.go
```diff
@@ -425,7 +429,8 @@ func runBuildLocal(ctx context.Context, cmd *cobra.Command, dst, spec string) {
 				SandboxTarget:     sandboxTarget,
 				// Only perform a build with the host DefaultPlatform at present.
 				// TODO: rework --arch handling for remote builds so that local builds can specify --arch and --platform.
-				Platform: *dp,
+				Platform:    *dp,
+				ReqAuthFile: reqAuthFile,
```
dtrudg, line 433: Perhaps, on thinking about it... this should be `DockerAuthFIle` in types.Options ... so it groups with the other things thare are specific to builds from docker sources?

I don't think we use it except for builds from docker sources?

PR #2252, opened 2023-10-11
add progress-bars for oras & ocisif operations

### LINKED ISSUE #2212: No progress bars in push/pull flows via go-containerregistry
In 4.0, some flows interacting with OCI registries now use go-containerregistry. This does not implement progress bars by default, and therefore these flows do not display download / upload progress bars.

We should ensure that progress bars are wired up to:

* ORAS push/pull - there have never been progress bars for this, even with the prior oras-go code, but there should be.
* Library OCI-SIF push/pull - I believe that lack of progress bars is a regression versus the native library API push/pull.

preminger:

Implements progress bars for:

- ORAS push/pull
- Library OCI-SIF push/pull

 - Fixes #2212

### DIFF
--- a/internal/pkg/client/ocisif/ocisif.go
+++ b/internal/pkg/client/ocisif/ocisif.go
@@ -295,6 +295,10 @@ func PushOCISIF(_ context.Context, sourceFile, destRef string, ociAuth *ocitypes
 				// The following is concurrency-safe because this is the only
 				// goroutine that's going to be reading progChan updates.
 				update := <-progChan
+				if update.Error != nil {
+					pb.Abort(false)
+					return
+				}
 				if update.Total != total {
 					pb.Init(update.Total)
 					total = update.Total
@@ -302,10 +306,10 @@ func PushOCISIF(_ context.Context, sourceFile, destRef string, ociAuth *ocitypes
 				pb.IncrBy(int(update.Complete - soFar))
 				soFar = update.Complete
 				if soFar >= total {
-					break
+					pb.Wait()
+					return
 				}
 			}
-			pb.Wait()
 		}()
 		remoteOpts = append(remoteOpts, remote.WithProgress(progChan))
 	}
--- a/internal/pkg/client/ocisif/ocisif.go
+++ b/internal/pkg/client/ocisif/ocisif.go
@@ -16,7 +16,7 @@ import (
 
 	ocitypes "github.com/containers/image/v5/types"
 	"github.com/google/go-containerregistry/pkg/name"
-	ggcr "github.com/google/go-containerregistry/pkg/v1"
+	ggcrv1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/empty"
 	"github.com/google/go-containerregistry/pkg/v1/layout"
 	ggcrmutate "github.com/google/go-containerregistry/pkg/v1/mutate"
@@ -46,7 +46,7 @@ type PullOptions struct {
 	DockerHost  string
 	NoHTTPS     bool
 	NoCleanUp   bool
-	Platform    ggcr.Platform
+	Platform    ggcrv1.Platform
 	ReqAuthFile string
 }
 
@@ -164,11 +164,11 @@ func createOciSif(ctx context.Context, sysCtx *ocitypes.SystemContext, imgCache 
 	if err != nil {
 		return err
 	}
-	digest, _, err := ggcr.SHA256(bytes.NewBuffer(rawManifest))
+	digest, _, err := ggcrv1.SHA256(bytes.NewBuffer(rawManifest))
 	if err != nil {
 		return err
 	}
-	mf, err := ggcr.ParseManifest(bytes.NewBuffer(rawManifest))
+	mf, err := ggcrv1.ParseManifest(bytes.NewBuffer(rawManifest))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


internal/pkg/client/ocisif/ocisif.go
```diff
@@ -16,7 +16,7 @@ import (
 
 	ocitypes "github.com/containers/image/v5/types"
 	"github.com/google/go-containerregistry/pkg/name"
-	v1 "github.com/google/go-containerregistry/pkg/v1"
+	ggcr "github.com/google/go-containerregistry/pkg/v1"
```
dtrudg, line 19: I'd prefer `ggcrv1` if you are going to change this - we have other packages from ggcr in here too, and the `v1` bit is important to know.

internal/pkg/ociimage/fetch.go
```diff
@@ -81,10 +82,14 @@ func FetchLayout(ctx context.Context, sysCtx *types.SystemContext, imgCache *cac
...
 	}
 
-	_, err = copy.Image(ctx, policyCtx, lr, srcRef, &copy.Options{
+	copyOpts := copy.Options{
 		ReportWriter: io.Discard,
 		SourceCtx:    sysCtx,
-	})
+	}
+	if term.IsTerminal(2) {
+		copyOpts.ReportWriter = os.Stdout
```
dtrudg, line 90: This needs to respect `--quiet` and `--silent` via looking for `sylog.GetLevel() <= -1` as the existing client.progress code does.

internal/pkg/client/ocisif/ocisif.go
```diff
@@ -282,5 +284,30 @@ func PushOCISIF(_ context.Context, sourceFile, destRef string, ociAuth *ocitypes
...
+				// goroutine that's going to be reading progChan updates.
+				update := <-progChan
+				if update.Total != total {
+					pb.Init(update.Total)
+					total = update.Total
+				}
+				pb.IncrBy(int(update.Complete - soFar))
+				soFar = update.Complete
+				if soFar >= total {
+					return
```
dtrudg, line 305: Should probably call `pb.Wait()` at this point, as that's the route to shutdown the progress bar / container go routines cleanly.

Note that Wait and Abort aren't seen called in the library flows as they called inside scs-library-client code.

internal/pkg/client/oras/oras.go
```diff
@@ -125,7 +128,32 @@ func UploadImage(_ context.Context, path, ref string, ociAuth *ocitypes.DockerAu
...
+				// goroutine that's going to be reading progChan updates.
+				update := <-progChan
+				if update.Total != total {
+					pb.Init(update.Total)
+					total = update.Total
+				}
+				pb.IncrBy(int(update.Complete - soFar))
+				soFar = update.Complete
+				if soFar >= total {
+					return
```
dtrudg, line 149: Should probably call `pb.Wait()` here, as above.

internal/pkg/client/oras/oras.go
```diff
@@ -17,21 +17,23 @@ import (
 
 	ocitypes "github.com/containers/image/v5/types"
 	"github.com/google/go-containerregistry/pkg/name"
-	v1 "github.com/google/go-containerregistry/pkg/v1"
+	ggcr "github.com/google/go-containerregistry/pkg/v1"
```
dtrudg, line 20: Ditto on `ggcrv1` naming. We have other ggcr packages in use.

internal/pkg/client/progress/progress_roundtripper.go
```diff
@@ -0,0 +1,43 @@
...
+		inner: inner,
+		pb:    pb,
+	}
+
+	return &rt
+}
+
+func (t *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.pb != nil && req.Body != nil && req.ContentLength >= contentSizeThreshold {
+		t.pb.Init(req.ContentLength)
```
dtrudg, line 34: Ideally every `Init` is paired with an `Abort` or `Wait` call when finished. A bit awkward here.... but if there's a way, then it should be added.

`mpb` implements multi-progress bar containers. It's expected that additional bars might be added, so the go routines associated with mpb don't exit until they are told to.

internal/pkg/ociimage/fetch.go
```diff
@@ -81,10 +82,14 @@ func FetchLayout(ctx context.Context, sysCtx *types.SystemContext, imgCache *cac
 		return nil, "", err
 	}
 
-	_, err = copy.Image(ctx, policyCtx, lr, srcRef, &copy.Options{
+	copyOpts := copy.Options{
 		ReportWriter: io.Discard,
 		SourceCtx:    sysCtx,
-	})
+	}
+	if term.IsTerminal(2) {
```
dtrudg, line 89: This needs to respect --quiet and --silent via looking for sylog.GetLevel() <= -1 as the existing client.progress code does.

dtrudg (changes requested):
Looks pretty good to me. There's no issue with having to tackle it differently in different areas of the code.

Just need to address missing handling of `--silent` `--quiet` in one place, and look into terminating the progress bard/container go routines cleanly where possible.

internal/pkg/ociimage/fetch.go
dtrudg: Whoops - duplicate. Sorry.

internal/pkg/client/ocisif/ocisif.go
preminger: Will do. The code that says `v1.Xyz` just looks very awkward to me - as in, it prompts the question "v1 of _what_?"

dtrudg (approved):
LGTM. Thanks!

PR #2302, opened 2023-11-02
fix: flatten overlay mounts in OCI launcher

dtrudg:

When a user specifies a read-only `--overlay` but no writable '--overlay', successful container startup requires that we add an ephemeral writable-tmpfs that runc / crun can write into.

Prior to this PR, '--overlay' overlays were handled and then the ephemeral writable-tmpfs was added in a separate overlayfs mount.

We can avoid 2 stacked overlayfs mounts by adding the ephemeral writable-tmpfs directly within 'WrapWithOverlays'.

This flattening is required so we don't exceed the 2 overlayfs stacking limit with multi-layer OCI-SIF images, which require an overlayfs mount to assemble the initial rootfs.

### DIFF
--- a/internal/pkg/runtime/launcher/oci/oci_overlay.go
+++ b/internal/pkg/runtime/launcher/oci/oci_overlay.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 
 	"github.com/sylabs/singularity/v4/internal/pkg/util/fs/overlay"
+	"github.com/sylabs/singularity/v4/pkg/image"
 	"github.com/sylabs/singularity/v4/pkg/ocibundle/tools"
 	"github.com/sylabs/singularity/v4/pkg/sylog"
 	"github.com/sylabs/singularity/v4/pkg/util/singularityconf"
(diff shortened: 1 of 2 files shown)


preminger (approved):
LGTM, except: I found it _very_ hard to keep track of what each method was doing when understanding that hinges on distinctions between "prepare" and "create" in the method name - I kept having to go back & read the contents of the method anew every time because it was very hard to figure out from the method names what is what.

If you have an idea of a better terminological distinction than "create" vs. "prepare", that would be great. (I say that being 95% sure I'm the one who introduced this terminology into this code, but that doesn't make it any better... 😉)

_ETA: Maybe "prepare" vs. "enact"...?_

dtrudg:
> LGTM, except: I found it _very_ hard to keep track of what each method was doing when understanding that hinges on distinctions between "prepare" and "create" in the method name - I kept having to go back & read the contents of the method anew every time because it was very hard to figure out from the method names what is what.
> 
> If you have an idea of a better terminological distinction than "create" vs. "prepare", that would be great. (I say that being 95% sure I'm the one who introduced this terminology into this code, but that doesn't make it any better... 😉)
> 
> _ETA: Maybe "prepare" vs. "enact"...?_

Noting that we've discussed this, and a 3-way split Prepare / Mount / PrepareAndMount... with Create being a deprecated alias for PrepareAndMount... seems a way forward.

I'll open an issue for that to be tackled separately.

Issue #2305, opened 2023-11-03
Impact of Compat default in OCI mode on User Experience


marcodelapierre:
**Is your feature request related to a problem? Please describe.**

Hey team, I have been deploying Singularity as the HPC container runtime as staff at the Pawsey national centre (AUS) for 4+ years. Now I am maintaining its interface with it within the workflow engine Nextflow as part of my new role at Seqera.
I have started to test the new OCI functionalities in V4. 
I am a bit concerned about `--oci` implying `--compat`, as it may disrupt user experience quite a bit. This is particularly true if a system admins decide to enable it by default in `singularity.conf`.

**Describe the solution you'd like**

I am raising the two following solutions for discussion:
1. adding a toggle in `singularity.conf` for `compat mode`, to enable sys admins to control the feature alongside `oci mode`; corresponding PR : https://github.com/sylabs/singularity/pull/2306
2. considering disabling the behaviour where `--oci` implies `--compat`, to minimise disruption of user experience.

Thoughts?

Thank you!

dtrudg:
Hi @marcodelapierre and thanks for opening this issue / discussion.

With the native runtime (i.e. SingularityCE 3.x or SingularityCE 4.x without `--oci`) the `--compat` flag was added to turn on features that bring behaviour as close as possible to Docker and other OCI runtimes, to fix some issues commonly experienced when people were directly running containers from Docker hub. The OCI compatibility that `--compat` can achieve is limited due to the nature of the native runtime and container format. For example, containers that use `USER` in their `Dockerfile` don't run exactly as under Docker.

The new `--oci` mode in SingularityCE 4.x is also explicitly about OCI compatibility, but now we use a modified OCI-SIF image format, and runc / crun as a true low-level runtime under the hood. It's our specific intention that `singularity run --oci ...` is very close to `docker run -it --rm ...` and _does not_ mirror historic Singularity behaviour.

> I am a bit concerned about `--oci` implying `--compat`, as it may disrupt user experience quite a bit. This is particularly true if a system admins decide to enable it by default in `singularity.conf`.

As above, the difference in behaviour is intentional. It would, in our view, be strange to have an `--oci` that implied `--no-compat` by default, since then it would be an `--oci` mode, that is by default not providing expected OCI container behaviour.

> I am raising the two following solutions for discussion:
> 
[...]

PR #2306, opened 2023-11-03
Adding singularity.conf option "compat mode"

marcodelapierre:

Hey team, I have been deploying Singularity as the HPC container runtime as staff at the Pawsey national centre (AUS) for 4+ years. Now I am maintaining its interface with it within the workflow engine Nextflow as part of my new role at Seqera.
I have started to test the new OCI functionalities in V4.
I am a bit concerned about `--oci` implying `--compat`, as it may disrupt user experience quite a bit. This is particularly true if a system admins decide to enable it by default in `singularity.conf`.

NOT done yet:
- Update User and Admin Docs
- [edit] Add unit tests (made an attempt in `config.go` but with no success) - help appreciated

 - Partially addresses https://github.com/sylabs/singularity/issues/2305

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
   `exec`, `shell`, and `instance start` can now also be passed a `--authfile
   <path>` option, to read OCI registry credentials from this custom file.
 
-- The `singularity.conf` configuration file now features a toggle to control 
+- The `singularity.conf` configuration file now features a toggle to control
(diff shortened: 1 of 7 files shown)


dtrudg (changes requested):
Hi @marcodelapierre - please could we discuss further on #2305 before you put any more effort into this?

I think this is unlikely to be merged as-is due to issues raised there.

marcodelapierre:
After reading your remarks in #2305, closing as I agree in full.

PR #2309, opened 2023-11-04
Update container-device-interface to v0.6.2

elezar:

Updated the CDI (container-orchestrated-devices) dependency to v0.6.2 which includes moving the CDI specs to a separately versioned go submodule. To do this I replaced the `github.com/container-orchestrated-devices/container-device-interface` imports and module entries with `tags.cncf.io/container-device-interface` and bumped version to `v0.6.2`.


dtrudg:
Many thanks for this. Should be good to go after CI completes. I just ran `scripts/update-license-dependencies.sh` to fix the failure. It's because there's a dependency bringing v.0.6.1 in indirectly from the old URI.

image

dtrudg (approved):

PR #2320, opened 2023-11-06
increase golangci-lint timeout to avoid sporadic CircleCI failures

preminger:

Increases the timeout for `golangci-lint` from `3m` to `5m` to avoid sporadic CircleCI failures.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
 run:
-  concurrency: 4
+  concurrency: 2
   timeout: 5m
   build-tags:
     - apparmor
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,6 @@
 run:
   concurrency: 4
-  timeout: 3m
+  timeout: 5m
   build-tags:
     - apparmor
     - containers_image_openpgp


dtrudg:
Ooof 5m for a meta-linter is pretty wild... our code isn't that big.

Is there any indication which specific linter is taking ages? I've been experiencing slowness in VSCode with linting updates which are a pain when doing large refactors... curious if we know why it is so slow?

dtrudg:
> > FWIW, the issue I was seeing with linting my code was that `golangci-lint` was pinning the CPU in the CircleCI instance causing it to be terminated. I ended up having to simply restrict the amount of CPU allocated to `golangci-lint` by setting the `--concurrency` to something less than the actual instance size.
> 
> We recently restricted `concurrency` to `4` to avoid a _different_ kind of CircleCI linter failure... I wonder if that somehow led to this 👆 ...?

That doesn't seem to be present here?

image

dtrudg:
> > FWIW, the issue I was seeing with linting my code was that `golangci-lint` was pinning the CPU in the CircleCI instance causing it to be terminated. I ended up having to simply restrict the amount of CPU allocated to `golangci-lint` by setting the `--concurrency` to something less than the actual instance size.
> 
> We recently restricted `concurrency` to `4` to avoid a _different_ kind of CircleCI linter failure... I wonder if that somehow led to this 👆 ...?



> > FWIW, the issue I was seeing with linting my code was that `golangci-lint` was pinning the CPU in the CircleCI instance causing it to be terminated. I ended up having to simply restrict the amount of CPU allocated to `golangci-lint` by setting the `--concurrency` to something less than the actual instance size.
> 
> We recently restricted `concurrency` to `4` to avoid a _different_ kind of CircleCI linter failure... I wonder if that somehow led to this 👆 ...?

What's the executor size for this task in CircleCI? We should likely match concurrency to the number of CPUs that are allocated.... and then anything else is going to be turning on/off linters that are making this take ages.... or upping the timeout :-/

preminger:
> What's the executor size for this task in CircleCI? We should likely match concurrency to the number of CPUs that are allocated.... and then anything else is going to be turning on/off linters that are making this take ages.... or upping the timeout :-/

This is all it says under the `executors:` section of .circleci/config.yml for golangci-lint:

```
  golangci-lint:
    docker:
      - image: golangci/golangci-lint:v1.55.1
```

EmmEff:
> What's the executor size for this task in CircleCI? We should likely match concurrency to the number of CPUs that are allocated.... and then anything else is going to be turning on/off linters that are making this take ages.... or upping the timeout :-/

Default Docker executor size in CircleCI is Medium (2 vCPUs/4GB RAM) which I believe also makes the `--concurrency=4` flag invalid (should really be `2` I think)

dtrudg:
> > What's the executor size for this task in CircleCI? We should likely match concurrency to the number of CPUs that are allocated.... and then anything else is going to be turning on/off linters that are making this take ages.... or upping the timeout :-/
> 
> Default Docker executor size in CircleCI is Medium (2 vCPUs/4GB RAM) which I believe also makes the `--concurrency=4` flag invalid (should really be `2` I think)

Thanks EmmEff - we're only using a non-default Large for the e2e-test step. Might be time to review executor size all-round, again, at some point.

dtrudg (approved):
Alright.

image

PR #2322, opened 2023-11-06
support multilayer Dockerfile build

preminger:

In Dockerfile build (using `buildkit`), honor `--keep-layers` command-line flag to create multilayer OCI-SIF image if requested.


dtrudg (changes requested):
We should probably error or warn if `--keep-layers` is provided for a non-OCI build, given there is no such thing as layers outside of OCI-SIF flows.

I probably need to check whether I did this in the pull / actions flows too.

PR #2328, opened 2023-11-07
fix: oci: use correct bind options for user /dev[/xxx] binds

### LINKED ISSUE #2326: oci: /dev/xxx -> /dev/xxx binds are performed with nodev flag
**Version of Singularity**

main

**Describe the bug**

In `--oci` mode, when a `/dev/xxx` -> `/dev/xxx` bind is performed, then `/dev/xxx` in the container is mounted with 'nodev'.

This stops e.g. `-B /dev/fuse` from providing a working FUSE device in the container. (Ref #2319)

In non-OCI native mode, there is logic that handles `/dev` -> `/dev` mounts so that they don't apply the `nodev` flag when performed to add devices to a minimal `/dev` that is provided by `--contain`.

dtrudg:

When a user requests a bind of all of `/dev`, or a single `/dev/xxx` device, into the container it is because they wish to use a device.

Prior to this PR, OCI-mode was binding `/dev` sources into the container with `nodev` set as a flag. Therefore e.g. `-B /dev/fuse` did not result in a usable FUSE device in the container.

Modify the code so it behaves like native mode. A user request to bind all of `/dev` is treated in the same way as the full `/dev` mount with `--no-compat`. A user request to bind a single `/dev/xxx` entry is performed with logic and flags for dev entries specifically.

This allows FUSE mounts inside OCI-Mode containers to work:

```
$ singularity run --keep-privs --oci -B .:/root -B /dev/fuse --fakeroot  docker://ubuntu
INFO:    Using cached OCI-SIF image
root@mini:~# ./gocryptfs cipher_oci plain_oci/
Password: 
Decrypting master key
DetectQuirks: Btrfs detected, forcing -noprealloc. See https://github.com/rfjakob/gocryptfs/issues/395 for why.
Filesystem mounted and ready.
```

 - Fixes #2326

### DIFF
--- a/internal/pkg/runtime/launcher/oci/mounts_linux.go
+++ b/internal/pkg/runtime/launcher/oci/mounts_linux.go
@@ -561,6 +561,11 @@ func (l *Launcher) addSystemBindMounts(mounts *[]specs.Mount) error {
 }
 
 func (l *Launcher) addUserBindMounts(mounts *[]specs.Mount) error {
+	if !l.singularityConf.UserBindControl {
+		sylog.Warningf("Ignoring bind mount request(s): user bind control disabled by system administrator")
+		return nil
+	}
+
 	// First get binds from -B/--bind and env var
 	binds, err := bind.ParseBindPath(strings.Join(l.cfg.BindPaths, ","))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/pkg/runtime/launcher/oci/mounts_linux.go
```diff
@@ -576,10 +581,18 @@ func (l *Launcher) addUserBindMounts(mounts *[]specs.Mount) error {
...
-		if !l.singularityConf.UserBindControl {
-			sylog.Warningf("Ignoring bind mount request: user bind control disabled by system administrator")
-			return nil
+		// Special Case - user is manually requesting all of /dev to be bound
+		if b.Source == "/dev" {
+			if b.Destination != "/dev" {
+				return fmt.Errorf("host /dev may only be bound to /dev in container")
+			}
+			// Handle as we do in --no-compat, including devpts setup.
+			if err := l.addDevBind(mounts); err != nil {
```
preminger, line 590: Should we be worried about circumstances (e.g. when --no-compat is passed) where addDevBind() is called twice - once from here, and once from addDevMounts()? Or do we want two mounts to be generated under that circumstance?

dtrudg: Hmmmm... I'm inclined to leave it as-is unless it causes a problem.

You can, for instance do someting like have `--no-compat` bind mount `/dev`.... a sysadmin has set `SINGULARITY_BINDS` to mount also a `/dev/mydevice` ... but you want to mask that by bind mounting `/dev` again over what's there.

Unlikely... probably... but I don't think allowing that is incorrect.

preminger (approved):

PR #2331, opened 2023-11-07
test: build from Dockerfile

preminger:

Adds test coverage for Dockerfile build functionality added in #2280 

Includes:

- [x] cross-arch builds
- [x] full suite of docker auth options
- [x] USER directive
- [x] use of already-running `builtkitd`
- [x] build args

### DIFF
--- a/e2e/docker/docker.go
+++ b/e2e/docker/docker.go
@@ -290,13 +290,14 @@ func (c ctx) testDockerCredsPriority(t *testing.T) {
 		}
 	})
 
-	dockerfile, err := e2e.WriteTempFile(tmpdir, "Dockerfile", fmt.Sprintf(
+	dockerfileContent := fmt.Sprintf(
 		`
 FROM %s
 CMD /bin/true
 `,
 		privImgNoPrefix,
-	))
+	)
+	dockerfile, err := e2e.WriteTempFile(tmpdir, "Dockerfile", dockerfileContent)
 	if err != nil {
 		t.Fatalf("while trying to generate test dockerfile: %v", err)
 	}
--- a/e2e/docker/docker.go
+++ b/e2e/docker/docker.go
@@ -1235,9 +1235,9 @@ func (c ctx) testDockerUSER(t *testing.T) {
 	)
 
 	metaTests := map[string]string{
-		"uri":  dockerURI,
-		"ubos": userBuiltOCISIF,
-		"rbos": rootBuiltOCISIF,
+		"uri":          dockerURI,
+		"user oci-sif": userBuiltOCISIF,
+		"root oci-sif": rootBuiltOCISIF,
 	}
 
 	for subtestName, container := range metaTests {
--- a/e2e/docker/docker.go
+++ b/e2e/docker/docker.go
@@ -17,6 +17,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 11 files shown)


e2e/imgbuild/imgbuild.go
```diff
@@ -2050,16 +2321,18 @@ func (c imgBuildTests) buildWithAuthTester(t *testing.T, withCustomAuthFile bool
...
-		`
-FROM %s
-CMD /bin/true
-`,
-		privImgNoPrefix,
-	))
-	if err != nil {
-		t.Fatalf("while trying to generate test dockerfile: %v", err)
-	}
+	// The following is disabled because it is now subsumed under
```
dtrudg, line 2324: Just remove this code. Significant commented chunks inevitably get out of date eventually, and can then confuse.

You can keep the removal in a separate commit with a message explaining it if you want to record the context of the removal.

e2e/imgbuild/imgbuild.go
```diff
@@ -2114,18 +2387,20 @@ CMD /bin/true
...
-			whileLoggedIn: false,
-			expectExit:    255,
-		},
-		{
-			name:          "privimg df logged in",
-			args:          []string{"-F", "--oci", "--no-https", "--disable-cache", "./my_image_file.oci.sif", dockerfile},
-			whileLoggedIn: true,
-			expectExit:    0,
-		},
+		// The following is disabled because it is now subsumed under
```
dtrudg, line 2390: Just remove this code. Significant commented chunks inevitably get out of date eventually, and can then confuse.

You can keep the removal in a separate commit with a message explaining it if you want to record the context of the removal.

test/defs/Dockerfile.buildargs
```diff
@@ -0,0 +1,8 @@
+FROM alpine
```
dtrudg, line 1: Apologies for not thinking about this before... but we try to ensure any tests consuming stuff from Docker hub are in `e2e/Docker`... so that it's very clear where things are that could contribute to hitting rate limits etc.

Would you be able to either:

* Move anything that's building `FROM <docker hub container>` into `e2e/docker` -or-
* Modify the definition files so that they build `FROM <our local registry image` (probably requires templating to substitute the correct FROM`.

e2e/docker/docker.go
```diff
@@ -1157,6 +1193,60 @@ func (c ctx) testDockerCMDQuotes(t *testing.T) {
...
+	}
+
+	for subtestName, whatToRun := range metaTests {
+		t.Run(subtestName, func(t *testing.T) {
+			c.testDockerUSERWorker(t, whatToRun)
+		})
+	}
+}
+
+func (c ctx) testDockerUSERWorker(t *testing.T, whatToRun string) {
```
dtrudg, line 1249: `whatToRun` is a bit ambiguous... consider e.g. whether `what` is the container I'm running stuff in, or `what` could be what to run inside a container.

I'd probably go for `container` or something like that.

e2e/docker/docker.go
```diff
@@ -1157,6 +1193,60 @@ func (c ctx) testDockerCMDQuotes(t *testing.T) {
...
+		e2e.AsSubtest("df build"),
+		e2e.WithProfile(e2e.UserProfile),
+		e2e.WithCommand("build"),
+		e2e.WithArgs("--oci", userBuiltOCISIF, dockerfile),
+		e2e.ExpectExit(0),
+	)
+	rootBuiltOCISIF := filepath.Join(tmpdir, "rootbuilt-docker-user.oci.sif")
+	c.env.RunSingularity(
+		t,
+		e2e.AsSubtest("df build"),
```
dtrudg, line 1215: Don't use the same subtest name twice. Go suffixes them with numbers then, and it's a bit confusing.

`user df build` and `root df build` should be fine.

e2e/docker/docker.go
```diff
@@ -1157,6 +1193,60 @@ func (c ctx) testDockerCMDQuotes(t *testing.T) {
...
+		e2e.ExpectExit(0, e2e.ExpectOutput(e2e.ContainMatch, fmt.Sprintf("uid=%d(%s) gid=%d",
+			e2e.UserProfile.ContainerUser(t).UID,
+			e2e.UserProfile.ContainerUser(t).Name,
+			e2e.UserProfile.ContainerUser(t).GID,
+		))),
+	)
+
+	metaTests := map[string]string{
+		"uri":  dockerURI,
+		"ubos": userBuiltOCISIF,
```
dtrudg, line 1238: ubos is getting a bit extreme with the shortening... "user oci-sif"?

e2e/docker/docker.go
```diff
@@ -271,6 +271,36 @@ func (c ctx) testDockerHost(t *testing.T) {
...
+	tmpdir, tmpdirCleanup := e2e.MakeTempDir(t, c.env.TestDir, "build-auth", "")
+	t.Cleanup(func() {
+		if !t.Failed() {
+			tmpdirCleanup(t)
+		}
+	})
+
+	dockerfile, err := e2e.WriteTempFile(tmpdir, "Dockerfile", fmt.Sprintf(
+		`
+FROM %s
```
dtrudg, line 294: If you declare this separately as a const and then use it in the `fmt.Sprintf` the code will be cleared to read. It's difficult to match brackets, commas etc. either side of a multi-line string literal.

e2e/imgbuild/imgbuild.go
```diff
@@ -1941,6 +1950,104 @@ DEMO=demo=with===equals==signs
...
+			tmpdirCleanup(t)
+		}
+	})
+
+	outputImgPath := filepath.Join(tmpdir, "image.oci.sif")
+
+	var cmd *exec.Cmd
+	var err error
+	var cmdPipe io.ReadCloser
+	e2e.Privileged(func(t *testing.T) {
```
dtrudg, line 1970: I don't think it's a great idea to launch a daemon as root from an e2e-test. If someone tries to run the test suite after building on a system where they may use buildkitd otherwise, we could cause issues.

I think it would be fair to run tests using one that's already running, or to start one rootless with a separate config file.

dtrudg (changes requested):

dtrudg (approved):
LGTM. Covers all the bases we need. Thanks!

PR #2342, opened 2023-11-09
build: more fixes to Dockerfiles build

preminger:

Two more fixes to Dockerfile builds (a.k.a. `buildkitd`-based builds):

1. Code in newSolveOpt() in internal/pkg/build/buildkit/client/client.go was not setting the `"context"` key of the `LocalDirs` field of `SolveOpt` correctly, causing `ADD` statements in Dockerfiles to fail. This PR fixes this (setting `LocalDirs["context"]` to the current working dir), and adds e2e code to verify it's working.
2. A leftover piece of debugging code hard-coded in internal/pkg/build/buildkit/client/client.go was disabling all build caching. This PR removes that code, and instead hooks up the toggling of `buildkitd`'s internal caching mechanism to the Singularity `--disable-cache` CLI option.

### DIFF
--- a/internal/pkg/build/buildkit/client/client.go
+++ b/internal/pkg/build/buildkit/client/client.go
@@ -72,6 +72,8 @@ type Opts struct {
 	KeepLayers bool
 	// Context dir in which to perform build (relevant for ADD statements, etc.)
 	ContextDir string
+	// Disable buildkitd's internal caching mechanism
+	DisableCache bool
 }
 
 func Run(ctx context.Context, opts *Opts, dest, spec string) {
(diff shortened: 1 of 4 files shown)


internal/pkg/build/buildkit/client/client.go
```diff
@@ -255,7 +257,7 @@ func newSolveOpt(_ context.Context, opts *Opts, w io.WriteCloser, buildDir, spec
 		spec = filepath.Join(buildDir, "Dockerfile")
 	}
 	localDirs := map[string]string{
-		"context":    buildDir,
+		"context":    "/",
```
dtrudg, line 260: This isn't correct for Dockerfile builds.

In Docker an `ADD` cannot add files outside of the context, and the context is that path specified on the command line.

See - https://docs.docker.com/build/building/context/

dtrudg (changes requested):

internal/pkg/build/buildkit/client/client.go
preminger: Right, ok – so we should probably discuss this. The previous value, `buildDir`, made it so that basically nothing could be ADDed, since it pointed to the temporary directory that didn't even exist prior to the build.

We could perhaps make it point to the current working directory by default, and perhaps wire up the `--cwd(/--pwd)` flag to the `build` command (in OCI mode only) to override this if the user wants to?

(Docker builds from Dockerfile, of course, typically involve an explicit directory argument, so this doesn't arise.)

dtrudg (approved):
LGTM

PR #2364, opened 2023-11-16
chore: prepare release 4.0.2

dtrudg:


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Changes Since Last Release
 
-### Changed defaults / behaviours
+## 4.0.2 \[2023-11-16\]
```
preminger, line 5: I might be misremembering this but... isn't it the "Changes Since Last Release" that should be changed to the version number, and the subdivision into "Changed defaults / behavio(u)rs" (😉) etc. should remain intact...?

preminger (changes requested):

CHANGELOG.md
dtrudg: Yep, you are right 🤦

preminger (approved):
LGTM!

PR #2374, opened 2023-11-17
document e2e helpers, best practices, etc.

### LINKED ISSUE #861: Document E2E helpers, best practice etc
**Type of issue**

Developer Documentation

**Description of issue**

The e2e suite is quite complicated, with a lot of helper functions, and established patterns for performing certain types of tests. These are not well documented, if at all.

We should have sufficient documentation and examples that a mid-level Go developer can read the e2e developer documentation and then be able to understand add new tests, or modify tests.

preminger:

Improve the documentation for the e2e testing suite: its structure, its initialization, the various helpers available, etc.

 - Fixes #861

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -314,8 +314,10 @@ Variables with the bolded names above are defined globally in e2e/internal/e2e,
 so that to access RootProfile from outside the e2e/internal/e2e package, for
 example, one would typically write `e2e.RootProfile`.
 
-However, e2e/internal/e2e/profile.go also defines three maps for convenient
-access to groups of profiles:
+### Convenience maps of profiles
+
+e2e/internal/e2e/profile.go also defines three maps for convenient access to
+groups of profiles:
 
 - NativeProfiles
 - OCIProfiles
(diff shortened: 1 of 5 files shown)


e2e/README.md
```diff
@@ -145,7 +149,1089 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
...
+
+- [`e2e.WriteTempFile(dir, pattern, content string) (string,
+  error)`](#temporary-dirs--files-and-cleanup)
+- [`e2e.MakeTempDir(t *testing.T, baseDir string, prefix string, context string)
+  (string, func(t *testing.T))`](#temporary-dirs--files-and-cleanup)
+
+Here are some additional utility functions that are available, and which are
+particularly useful for writing e2e tests:
+
+- The `require` package (internal/pkg/test/tool/require/require.go)
```
dtrudg, line 1073: Ideally add a link to the file in GitHub main branch.

e2e/README.md
```diff
@@ -145,7 +149,1089 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
...
+    templates](https://pkg.go.dev/text/template) in tests.
+    - Like the `require` package, it is available for unit-tests as well as e2e
+      tests.
+  - Execute() injects a given set of `values` into a template (whose path is
+    given by `tmplPath`), and places the result in a new, temporary file created
+    in `tmpdir` and named using the pattern `namePattern`.
+  - The created file is automatically removed at the end of the test `t`, unless
+    the test fails.
+
+## Common pitfalls
```
dtrudg, line 1115: Hopefully we'lll continue to add things here, so it'd be good if each of these scenarios had a concise heading, so it's easy to scan through them as the list grows.

E.g. 

'Test not visible in logs'

'E2E_GROUPS / E2E_TESTS filtering not working'

e2e/README.md
```diff
@@ -145,7 +149,1089 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
...
+```
+
+`TestE2E` is the single top-level name for all e2e tests; next comes [`PAR` or
+`SEQ`](#parallel-par-vs-non-parallel-seq-tests), then comes the group name
+(which is what E2E_GROUPS matches), followed by the top-level test name (which
+is what E2E_TESTS matches; in this case, `auth`). The E2E_TESTS environment
+variable is a regular expression that is matched only against the top-level test
+name.
+
+At the present, there is no way to match against more deeply-embedded components
```
dtrudg, line 1153: You can actually do this with `make -C builddir e2e-test -run /my/specific/test/name/here`

e2e/README.md
```diff
@@ -145,7 +149,1089 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
...
+		"use":            c.remoteUse,
+		"use exclusive":  np(c.remoteUseExclusive),
+	}
+}
+```
+
+To run only the c.remoteUseExclusive() test, the environment variable value
+`E2E_TESTS=remoteUseExclusive` won't work. That's because, as far as the testing
+suite is concerned, the test's name is `use exclusive`. Therefore, a correct
+value to run this test only would be, e.g., `E2E_TESTS=use.exclusive`.
```
dtrudg, line 1187:
```suggestion
value to run this test only would be, e.g., `E2E_TESTS=use exclusive`.
```

e2e/README.md
```diff
@@ -145,7 +149,1089 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
...
+the file e2e/imgbuild/imgbuild.go, which contains the code for the package
+`imgbuild`, is responsible for the e2e group named `BUILD` (and *not*
+`IMGBUILD`!). This can be seen by consulting the `e2eGroups` map defined in
+e2e/suite.go:
+
+```go
+var e2eGroups = map[string]testhelper.Group{
+	"ACTIONS":    actions.E2ETests,
+	"BUILDCFG":   e2ebuildcfg.E2ETests,
+	"BUILD":      imgbuild.E2ETests, // <--- this one!
```
dtrudg, line 1200: There's no real reason not to change this at this point, I don't think. Having to point it out as a gotcha (I think ENV is too) probably shows we should fix it...

e2e/internal/e2e/image.go
```diff
@@ -200,29 +200,6 @@ func EnsureORASImage(t *testing.T, env TestEnv) {
 	})
 }
 
-var orasPrivImageOnce sync.Once
-
-func EnsureORASPrivImage(t *testing.T, env TestEnv) {
```
dtrudg, line 205: Please don't delete code in a docs specific PR.

dtrudg (changes requested):
Looks pretty good to me. Thanks for the efforts on this, which will definitely assist contributors in future.

There are some specific technical comments, but most things are nit-pick comments regarding vocabulary - that you are free to ignore. However, try to aim for the most simple & concise language possible in these types of documents.

Generally we cannot assume knowledge of terms like 'variadic' versus 'zero or more'. Historically, a large portion of the contributors to Singularity have been self-taught programmers, who may not be aware of the specific meaning of more technical terms.

e2e/README.md
```diff
@@ -145,7 +149,1089 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
...
+
+The first argument is the Unix [exit
+code](https://www.baeldung.com/linux/status-codes) the test expects the CLI to
+return. (Zero, as in this example, means that the CLI run has terminated
+successfully; though that does not yet mean it did what we expected; keep
+reading!)
+
+The rest of the arguments to e2e.ExpectExit() are a variadic set of [functional
+options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)
+again. *(Yes: functional options inside a functional option!)* The most common
```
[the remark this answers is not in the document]
preminger: Boo. :-(

e2e/README.md
```diff
@@ -145,7 +149,1089 @@ func E2ETests(env e2e.TestEnv) testhelper.Tests {
...
+<...>
+TestE2E/SEQ/DOCKER/cred_prio
+<...>
+```
+
+`TestE2E` is the test name for the entire e2e suite; it is followed by `PAR`,
+for the set of tests run in parallel, or by `SEQ`, for those tests that cannot
+be run in parallel and are run sequentially.
+
+The fact that the testhelper.NoParallel() function returns its argument as its
```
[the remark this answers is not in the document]
preminger: I've simplified the first half of the explanation (as per your previous comment). But I think the part about the return value is important, otherwise the way `testhelper.Tests` maps get written doesn't really make sense. If `Fn` is a function that "marks its argument as X", you would usually conclude from that that the way to use `Fn` is to call `Fn(arg)` (with no return value or an error return value), and then use `arg` when you need its value. `testhelper.NoParallel` is written in a way that this will also work, but I think that in a README, we want to encourage people to use the pattern of marking a non-parallel test "in place" —

```go
		"test help":      c.remoteTestHelp,
		"use":            c.remoteUse,
		"use exclusive":  np(c.remoteUseExclusive),
```

e2e/README.md
preminger: Ooh good idea!

e2e/README.md
preminger: What's wrong with using a dot here...? Note that, without the dot, you'd have to quote the value in some contexts (e.g. when using the prefixed command line pattern, as in `E2E_TESTS=use.exclusive make -C ...`, whereas with the dot, you don't.

e2e/README.md
preminger: It's not just these two... it's also "BUILDCFG" vs. "e2ebuildcfg". I wasn't sure there whether there was some reason which I wasn't aware of behind these choices. In any event, that's a significant change in the code; I'd lean towards making that a separate PR, that then also includes removing this part of the discussion here. What do you think?

preminger: We could also fold the zapping of EnsureORASPrivImage() into that same PR. (The issue there is not just that it's unused code; there's a fundamental problem with the idea behind it, and in any event it's covered by our testing in a different way now.)

e2e/README.md
dtrudg: Using the `.` without explaining how it works is inviting people to think that e.g. Go replaces spaces in test names with `.`

I'd prefer showing the quoting, as that's correct for matching only that test name. Otherwise please mention why the `.` works.

e2e/README.md
dtrudg: Yeah, sorry - I was assuming the changes could/would be made in a PR merged ahead of this one. I should have been clear.

If the `EnsureORASPrivImage` is going to be removed, let's have a completely separate PR for that. I'm not clear whether it's a clean backport, and to which versions.... separate PR makes it easier to reason about that.

dtrudg (approved):

Issue #1022, opened 2022-10-10
Add --oci flag to CLI that calls OCI launcher instead of singularity starter based launcher.


dtrudg:
Add a new flag `--oci` for actions (run/exec/shell) in the CLI layer. This flag will instruct the CLI to call the new OCI launcher from #597 to launch the container, instead of the singularity runtime starter launcher.

The launcher is not functional, and using `--oci` should simply give an error at this stage.

PR #2376, opened 2023-11-20
refactor: minimise containers/image imports

dtrudg:

Refactor code so that we minimise the number of times that we import from `containers/image`, without yet rewriting the image pulling & caching code that uses `containers/image`.

Use the ggcr AuthConfig over the containers/image DockerAuthconfig.

Use our own TransportOptions over the containers/image SystemContext.

Where we still have functional containers/image code, translate to its types as near as possible.

A first step for #2022 


cmd/internal/cli/actions.go
```diff
@@ -185,7 +185,7 @@ func replaceURIWithImage(ctx context.Context, cmd *cobra.Command, args []string)
 		image, err = handleOras(ctx, imgCache, cmd, origImageURI)
 	case uri.Shub:
 		image, err = handleShub(ctx, imgCache, origImageURI)
-	case oci.IsSupported(t):
+	case ocitransport.SupportedTransport(t):
```
preminger, line 188: Given the package name (which I assume is intended to stay `ocitransport`, would it make sense to keep this as IsSupported() (or Supported()) so that it doesn't stutter?

cmd/internal/cli/build.go
```diff
@@ -506,34 +506,35 @@ func definitionFromSpec(spec string) (types.Definition, error) {
 	return def, nil
 }
 
-// makeDockerCredentials creates an *ocitypes.DockerAuthConfig to use for
-// OCI/Docker registry operation configuration. Note that if we don't have a
-// username or password set it will return a nil pointer, as containers/image
-// requires this to fall back to .docker/config based authentication.
-func makeDockerCredentials(cmd *cobra.Command) (authConf *ocitypes.DockerAuthConfig, err error) {
+// makeOCICredentials creates an *authn.AuthConfig to use for OCI/Docker
+// registry operation configuration. Note that if we don't have a username or
```
preminger, line 510: Given that this function calls out to interactive login under certain circumstances, would it make sense to rewrite the part of the comment beginning with "Note that if" (probably written before interactive login was even implemented...)?

internal/pkg/build/sources/conveyorPacker_oci.go
```diff
@@ -110,35 +112,32 @@ exec "$@"
...
+		Insecure:         cp.b.Opts.NoHTTPS,
+		DockerDaemonHost: cp.b.Opts.DockerDaemonHost,
+		AuthConfig:       cp.b.Opts.OCIAuthConfig,
+		AuthFilePath:     ociauth.ChooseAuthFile(cp.b.Opts.DockerAuthFile),
+		UserAgent:        useragent.Value(),
+		TmpDir:           b.TmpDir,
+		Platform:         cp.b.Opts.Platform,
+	}
+
+	if cp.b.Opts.OCIAuthConfig == nil && cp.b.Opts.DockerAuthConfig != nil {
```
preminger, line 135: I'm *guessing* that the reason `cp.b.Opts` now has both an `OCIAuthConfig` field and a `DockerAuthConfig` (as opposed to unifying them into a single `OCIAuthConfig`, as has been done elsewhere) is that `pkg/build/types` is under the `pkg/` toplevel. Can you confirm?

internal/pkg/build/sources/conveyorPacker_oci.go
```diff
@@ -211,7 +210,9 @@ func (cp *OCIConveyorPacker) Pack(ctx context.Context) (*sytypes.Bundle, error)
 }
 
 func (cp *OCIConveyorPacker) getConfig(ctx context.Context) (imgspecv1.ImageConfig, error) {
-	img, err := cp.srcRef.NewImage(ctx, cp.sysCtx)
+	// TODO - replace with ggcr code
+	//nolint:staticcheck
+	img, err := cp.srcRef.NewImage(ctx, ocitransport.SystemContextFromTransportOptions(cp.transportOptions))
```
preminger, line 215: As above, I assume these are remnants of the `containers/image` API that can't be excised until the next major version (at the earliest)...?

internal/pkg/build/sources/conveyorPacker_oci.go
```diff
@@ -235,7 +236,14 @@ func (cp *OCIConveyorPacker) insertOCIConfig() error {
 }
 
 func (cp *OCIConveyorPacker) unpackTmpfs(ctx context.Context) error {
-	imageSource, err := cp.srcRef.NewImageSource(ctx, cp.sysCtx)
+	var sysCtx *types.SystemContext
+	if cp.transportOptions != nil {
```
preminger, line 240: I'm confused about something: why not use ocitransport.SystemContextFromTransportOptions() here, as in line 215? (I think it even includes the same `nil` check that is done explicitly here.)

internal/pkg/client/library/registry.go
```diff
@@ -80,13 +80,13 @@ func (lr *libraryRegistry) convertRef(libraryRef scslibrary.Ref) (string, error)
 	return ref, nil
 }
 
-// authConfig returns a DockerAuthConfig with current token to authenticate
+// authConfig returns a authn.AuthConfig with current token to authenticate
```
preminger, line 83: Nit of all nits:

```suggestion
// authConfig returns an authn.AuthConfig with current token to authenticate
```

internal/pkg/client/oci/nativesif.go
```diff
@@ -13,23 +13,21 @@ import (
...
 	"github.com/sylabs/singularity/v4/pkg/sylog"
 )
 
 // pullNativeSIF will build a SIF image into the cache if directTo="", or a specific file if directTo is set.
 func pullNativeSIF(ctx context.Context, imgCache *cache.Handle, directTo, pullFrom string, opts PullOptions) (imagePath string, err error) {
-	sys, err := sysCtx(opts)
-	if err != nil {
-		return "", err
-	}
+	to := transportOptions(opts)
```
preminger, line 23: Another nit: I personally don't love `to` as the name for a variable. I realize it's not a keyword in Go, but it is very keyword-*ish*, if that makes sense. It's really hard to read it as anything other than the English preposition "to." How about something like `tOpts`?

(Comment applies in numerous other places in the PR, but I didn't want to pollute the review with repetitions of this, so just do a search if you decide to implement.)

internal/pkg/client/oci/nativesif.go
```diff
@@ -75,20 +73,20 @@ func convertOciToSIF(ctx context.Context, imgCache *cache.Handle, image, cachedI
...
 			Opts: buildtypes.Options{
 				TmpDir:           opts.TmpDir,
 				NoCache:          imgCache.IsDisabled(),
 				NoTest:           true,
 				NoHTTPS:          opts.NoHTTPS,
-				DockerAuthConfig: opts.OciAuth,
 				DockerDaemonHost: opts.DockerHost,
 				ImgCache:         imgCache,
 				Platform:         opts.Platform,
+				OCIAuthConfig:    opts.OciAuth,
```
preminger, line 86: If we've renamed this from `DockerAuthXYZ` to `OCIAuthXYZ`, why not rename `DockerAuthFile` (next line) to `OCIAuthFile` as well?

preminger (approved):
Looks good to me; left bunch of comments, all of them stylistic.

cmd/internal/cli/actions.go
dtrudg: I'm going to leave this as-is here... the `ocitransport` package is a temporary thing to break an import cycle that otherwise occurs across client and build code. It should go away as the #1022 work continues.

internal/pkg/build/sources/conveyorPacker_oci.go
dtrudg: Yes - we have to keep `DockerAuthConfig` there, and working until v5.

internal/pkg/build/sources/conveyorPacker_oci.go
dtrudg: Not quite... this will be rewritten into ggcr code within later PRs for 4.1 without any change to the public API / behaviour. I didn't want to make this PR much larger by doing that in one step.

internal/pkg/client/oci/nativesif.go
dtrudg: Yeah, probably fair. Especially when we're in the realm of moving imaged from somewhere, to somewhere.

internal/pkg/client/oci/nativesif.go
dtrudg: The auth file is _not_ an OCI standard. The format of this file is specifically something that is defined by Docker's implementation, therefore it's a DockerAuthFile.

dtrudg: To be fair - the AuthConfig isn't an OCI standard either... but is is something that applies to accessing OCI registries of all kinds.

PR #2661, opened 2024-02-15
fix: oci: always use rootless uid for inner uid map

### LINKED ISSUE #2623: oci: container USER is affecting inner uid_map
**Version of Singularity**

main / 4.1

**Describe the bug**

At some point, the `USER` uid, picked up from an image config has started to addect the inner uid_map of OCI-Mode rootless.

The map is being calculated based on the `USER` uid, and not the uid of the person starting the container. This will cause incorrect ownership of files in the container image.

We get:

```
$ singularity shell --oci docker://sylabsio/docker-user
INFO:    Using cached OCI-SIF image
Singularity> cat /proc/self/uid_map
         0          1       2000
      2000          0          1
      2001       2001      63536
Singularity> 
```

We should have:

```
~ $ cat /proc/self/uid_map 
         0          1       1000
      1000          0          1
      1001       1001      64536
```

dtrudg:

In OCI mode, as an unprivileged user, we have an outer UID/GID mapping from the host UID/GID to 0:0 and an inner UID/GID mapping back to the host UID/GID.

The code was erroneously using the targetUID/GID for the container process rather than the host UID/GID for the inner mapping. This meant that when a container specifying an Dockerfile `USER` was run the mapping and file owneship in the container was incorrect.

Fix the mapping and add e2e-tests that verify correct file ownership for a container with a `USER`.

 - Fixes #2623

### DIFF
--- a/internal/pkg/runtime/launcher/oci/launcher_linux.go
+++ b/internal/pkg/runtime/launcher/oci/launcher_linux.go
@@ -320,7 +320,7 @@ func (l *Launcher) finalizeSpec(ctx context.Context, b ocibundle.Bundle, spec *s
 	containerUser := false
 
 	// If the OCI image config specifies a USER we will:
-	//  * When unprivileged - run as that user, via nested subuid/gid mappings (host user -> userns root -> OCI USER)
+	//  * When unprivileged - run as that user, via nested subuid/gid mappings.
 	//  * When privileged - directly run as that user, as a host uid/gid.
 	if imgSpec.Config.User != "" {
 		imgUser, err := tools.BundleUser(b.Path(), imgSpec.Config.User)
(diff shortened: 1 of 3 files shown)


internal/pkg/runtime/launcher/oci/launcher_linux.go
```diff
@@ -320,7 +320,7 @@ func (l *Launcher) finalizeSpec(ctx context.Context, b ocibundle.Bundle, spec *s
 	containerUser := false
 
 	// If the OCI image config specifies a USER we will:
-	//  * When unprivileged - run as that user, via nested subuid/gid mappings (host user -> userns root -> OCI USER)
```
dtrudg, line 323: Note - this comment wasn't really incorrect... where the parenthesized portion refers to the effective UID at each stage.

However it's easy to read it as if the parenthesized portion describes the ID mapping.. which may have led to this bug.

EmmEff (approved):

PR #2767, opened 2024-03-21
Bump go to minimum of 1.21

dtrudg:

Update the minimum version of Go to 1.21 so we can bring in dependencies that declare the same.

Doing this as a separate PR as it requires a fix in code using `unsafe.` for a deprecation to fix a linter failure:

chore: fix unsafe.StringHeader lint

If we are running with a PID namespace, we have a shim process by default. We have code that sets the name of this process to `sinit` by:

 * Overwriting the value of argv[0]
 * Using PR_SET_NAME

Both are necessary as PR_SET_NAME only affects the value from PR_GET_NAME.

In this code, unsafe.StringHeader is now deprecated. Rewrite using the pattern that takes an unsafe.Slice from an unsafe.StringData to get the raw bytes for the string os.Args[0].

See:

https://github.com/golang/go/issues/53003#issuecomment-1964775055


internal/pkg/runtime/engine/singularity/process_linux.go
```diff
@@ -220,17 +219,19 @@ func (e *EngineOperations) StartProcess(masterConn net.Conn) error {
 	// Modify argv argument and program name shown in /proc/self/comm
 	name := "sinit"
 
-	argv0str := (*reflect.StringHeader)(unsafe.Pointer(&os.Args[0]))
-	argv0 := (*[1 << 30]byte)(unsafe.Pointer(argv0str.Data))[:argv0str.Len]
-	progname := make([]byte, argv0str.Len)
+	argv0 := unsafe.Slice(unsafe.StringData(os.Args[0]), len(os.Args[0]))
```
dtrudg, line 222: For some context around the change here, see: https://github.com/golang/go/issues/53003#issuecomment-1964775055

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/sylabs/singularity/v4
 
-go 1.20
+go 1.21
+
+toolchain go1.21.8
```
tri-adam, line 5: If I understand correctly, the `toolchain` line will cause an installed version of Go between 1.21.0 and 1.21.7 to download and use 1.21.8 via automatic toolchain switching. Is that the intent here? And if so, will we plan to bump this each time the Go team releases a patch release?

Haven't seen use of this in the wild yet, so not sure what the best practice is!

tri-adam (approved):
LGTM, just one question, thanks!

go.mod
dtrudg: > If I understand correctly, the `toolchain` line will cause an installed version of Go between 1.21.0 and 1.21.7 to download and use 1.21.8 via automatic toolchain switching. Is that the intent here? 

That's my understanding of the behaviour, yes.

I included it here, because on dependabot updates our go.mod check is complaining about it missing... see e.g. the dependabot PRs open for the CNI plugins:

image

> And if so, will we plan to bump this each time the Go team releases a patch release?
> Haven't seen use of this in the wild yet, so not sure what the best practice is!

I'm a bit unclear on this tbh. My gut feeling is that for projects like ours, it is useful to inform a minimum Go toolchain version to avoid a relevant security issue. Therefore I'd anticipate we'd bump it if there's some security issue in a Go release that's relevant for Singularity?

We could drop the toolchain line for now... and then work around anything that comes up on the check go.mod task?

dtrudg: After some discussion... will drop the toolchain line at this time, to avoid issues with distros etc. where fetching a toolchain would be frowned upon.

PR #2837, opened 2024-04-19
fix: starter: overwrite glibc's internal tid cache on clone()

### LINKED ISSUE #2677: SIGSEV (signal 11 error) with Go 1.22.0 and Ubuntu 20.04 / Debian 10
**Version of Singularity**

Singularity-CE 4.1.0
(also tried CE 4.1.1 with same result)

**Describe the bug**

Fails to mount tempfs or ramfs for "build --sandbox" mainly with this error: signal number 11

"VERBOSE [U=0,P=1]          wait_child()                  rpc server interrupted by signal number 11
FATAL   [U=0,P=9662]       Master()                      container creation failed: mount tmpfs->/usr/local/var/singularity/mnt/session error: while mounting tmpfs: can't mount tmpfs filesystem to /usr/local/var/singularity/mnt/session: read unix @->@: read: connection reset by peer
: exit status 255"

**To Reproduce**

Reset my dual-boot Linux harddrive back to its initial conditions for Ubuntu focal 20.04.6 LTS
installed a small number of programs needed for using the image to be built from my custom definition
full list of commands / packages in attached file:
[full_reproduction.odt](https://github.com/sylabs/singularity/files/14395771/full_reproduction.odt)

Install GO 1.22.0 
Install dependencies listed at (https://docs.sylabs.io/guides/main/admin-guide/installation.html) for Ubuntu
Install singularity into usr/local with:

./mconfig && \
    make -C ./builddir && \
    sudo make -C ./builddir install

Run build -- sandbox on my custom definition file with a library for ubuntu:20.04 as its base
AS WELL AS on this basic test case library:
**sudo singularity -d build --sandbox ubuntu/ library://ubuntu**
Fails to mount with above error

Tried more by setting these env variables & with or without various config file settings listed
With or Without these env variables
export SINGULARITY_TMPDIR=/home/giovannini/sandbox/temp/tmp
export SINGULARITY_CACHEDIR=/home/giovannini/sandbox/temp/cache
		(also tried with or without sudo -E)

	(also tried with singularity_conf	mount tmp = no or yes)
	(also tried with singularity_conf	mount host fs = no or yes)
	(also tried with singularity_conf	sessiondir_max_size = 20480 or 64)

Also tried with "mount fs" set to tempfs and ramfs.
All fail to mount at the same point.

**Expected behavior**

Expected to mount tempfs/ramfs to create a sandbox folder OR the .sif from my defintion

**OS / Linux Distribution**

Ubuntu focal 20.04.6 LTS

**Installation Method**

Install singularity into usr/local with:
./mconfig && \
    make -C ./builddir && \
    sudo make -C ./builddir install
using your github release source file for 4.1.0 (and previously 4.1.1 before resetting my OS)

**Additional context**

mount and cat /proc/self/mountinfo 
and build config also in attached file:
[mount_build_info.odt](https://github.com/sylabs/singularity/files/14395777/mount_build_info.odt)

DEBUG
Example Debug when attempting to sandbox my definition file.
Same point of failure for the other basic test case.
	with or without these variables set
export SINGULARITY_TMPDIR=/home/giovannini/sandbox/temp/tmp
export SINGULARITY_CACHEDIR=/home/giovannini/sandbox/temp/cache
	
	(also tried with singularity_conf	mount tmp = no or yes)
	(also tried with singularity_conf	mount host fs = no or yes)
	(also tried with singularity_conf	sessiondir_max_size = 20480 or 64)
	(also tried with or without sudo -E for those env vars)

[Debug Log.odt](https://github.com/sylabs/singularity/files/14395789/Debug.Log.odt)

dtrudg:
Adapted from: https://github.com/opencontainers/runc/pull/4247

Execution of a container using a PID namespace can fail on certain versions of glibc when Singularity is built with Go 1.22.

This is due to Go 1.22 performing calls using pthread_self which, from glibc 2.25, is not updated for the current TID on clone.

Fixes #2677

-----

Original runc explanation:

Since glibc 2.25, the thread-local cache of the current TID is no longer updated in the child when calling clone(2). This results in very unfortunate behaviour when Go does pthread calls using pthread_self(), which has the wrong TID stored.

The "simple" solution is to forcefully overwrite this cached value. Unfortunately (and unsurprisingly), the layout of "struct pthread" is strictly private and can change without warning.

Luckily, glibc (currently) uses CLONE_CHILD_CLEARTID for all forks (with the child_tid set to the cached &PTHREAD_SELF->tid), meaning that as long as runc is using glibc, when "runc init" is spawned the child process will have a pointer directly to the cached value we want to change. With CONFIG_CHECKPOINT_RESTORE=y kernels on Linux 3.5 and later, we can simply use prctl(PR_GET_TID_ADDRESS). For older kernels we need to memory scan the TLS structure (pthread_self() returns a pointer to the start of the structure so we can "just" scan it for a field containing the current TID and assume that it is the correct field).

Obviously this is all very horrific, and if you are reading this in the future, it almost certainly has caused some horrific bug that I did not forsee. Sorry about that. As far as I can tell, there is no other workable solution that doesn't also depend on the CLONE_CHILD_CLEARTID behaviour of glibc in some way. We cannot "just" do a re-exec after clone(2) for security reasons.

Fixes https://github.com/opencontainers/runc/issues/4233 Signed-off-by: Aleksa Sarai cyphar@cyphar.com

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ orbs:
 parameters:
   go-version:
     type: string
-    default: '1.21.9'
+    default: '1.22.2'
 
 executors:
   node:
--- a/LICENSE_THIRD_PARTY.md
+++ b/LICENSE_THIRD_PARTY.md
@@ -607,3 +607,31 @@ Contain code from the docker cli project, under the Apache License, Version 2.0.
    limitations under the License.
 */
 ```
+
+## github.com/containers/runc
+
+The source file:
+
+* `cmd/starter/c/starter.c`
+
+Contains code from the runc project, under the Apache License, Version 2.0.
+
+``` text
+runc
+
+Copyright 2012-2015 Docker, Inc.
+
+This product includes software developed at Docker, Inc. (http://www.docker.com).
+
+The following is courtesy of our legal counsel:
+
+
+Use and transfer of Docker may be subject to certain restrictions by the
+United States and other governments.
+It is your responsibility to ensure that your use and/or transfer does not
+violate applicable laws.
+
+For more information, please see http://www.bis.doc.gov
+
+See also http://www.apache.org/dev/crypto.html and/or seek legal counsel.
+```
--- a/cmd/starter/c/starter.c
+++ b/cmd/starter/c/starter.c
@@ -1,11 +1,28 @@
 /*
-  Copyright (c) 2018-2023, Sylabs, Inc. All rights reserved.
+  Copyright (c) 2018-2024, Sylabs, Inc. All rights reserved.
 
   This software is licensed under a 3-clause BSD license.  Please
   consult LICENSE.md file distributed with the sources of this project regarding
   your rights to use or distribute this software.
-*/
 
+  Portions of this code are adapted from:
+    https://github.com/opencontainers/runc/pull/4247
+
+  Copyright 2012-2015 Docker, Inc.
+
+  This product includes software developed at Docker, Inc. (http://www.docker.com).
+
+  The following is courtesy of our legal counsel:
+
+  Use and transfer of Docker may be subject to certain restrictions by the
+  United States and other governments.
+  It is your responsibility to ensure that your use and/or transfer does not
+  violate applicable laws.
+
+  For more information, please see http://www.bis.doc.gov
+
+  See also http://www.apache.org/dev/crypto.html and/or seek legal counsel.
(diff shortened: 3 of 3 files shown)


dtrudg:
Shouldn't be needed after https://go-review.googlesource.com/c/go/+/587919 is backported, hopefully into 1.22.4.

PR #2855, opened 2024-04-30
Allows using multiple --env-file flags.

### LINKED ISSUE #2744: support passing multiple env files

I would like to be able to pass environmental variables that are stored in separate files, eg:

```bash
            singularity exec --bind .:/app \
                             --env-file ./.envs/.local/.django \
                             --env-file ./.envs/.local/.postgres \
                             --env-file ./.envs/.local/.regulatory_data \
                             $django_sif bash -c 'cd /app && /start'
```

The environmental variables from each file should be set in the container environment, with later files taking priority

Keeping the environmental variables in separate files can be useful, but of course this can be concatenated and provided to `--env-file` as it is currently implemented

cyanezstange:

Write your description of the PR here. Be sure to include as much background,
and details necessary for the reviewers to understand exactly what this is
fixing or enhancing.

 - Fixes #2744

### DIFF
--- a/internal/pkg/runtime/launcher/native/launcher_linux.go
+++ b/internal/pkg/runtime/launcher/native/launcher_linux.go
@@ -882,7 +882,7 @@ func (l *Launcher) setEnv(ctx context.Context, args []string) error {
 		)
 
 		// Read all environment files and put the variables into envFilesMap,
-		// environment variables in latter files will take precedence.
+		// environment variables in later files will take precedence.
 		envFilesMap := map[string]string{}
 		for _, envFile := range l.cfg.EnvFiles {
 			tempEnvMap, err := env.FileMap(ctx, envFile, args, currentEnv)
--- a/internal/pkg/runtime/launcher/native/launcher_linux.go
+++ b/internal/pkg/runtime/launcher/native/launcher_linux.go
@@ -875,27 +875,30 @@ func (l *Launcher) setNamespaces() {
 
 // setEnv sets the environment for the container, from the host environment, glads, env-file.
 func (l *Launcher) setEnv(ctx context.Context, args []string) error {
-	if l.cfg.EnvFile != "" {
+	if len(l.cfg.EnvFiles) > 0 {
 		currentEnv := append(
 			os.Environ(),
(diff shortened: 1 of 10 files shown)


internal/pkg/runtime/launcher/native/launcher_linux.go
```diff
@@ -875,25 +875,29 @@ func (l *Launcher) setNamespaces() {
...
-			return fmt.Errorf("while processing %s: %w", l.cfg.EnvFile, err)
+		// Read all environment files and put the variables into envFilesMap,
+		// environment variables in later files will take precedence.
+		envFilesMap := map[string]string{}
+		for _, envFile := range l.cfg.EnvFiles {
+			tempEnvMap, err := env.FileMap(ctx, envFile, args, currentEnv)
+			if err != nil {
+				return fmt.Errorf("while processing %s: %w", envFile, err)
+			}
+			envFilesMap = env.MergeEnvFileMap(envFilesMap, tempEnvMap, envFile)
```
dtrudg, line 892: It's a little bit unusual to pass the filename `envFile` here, when it's only used for log messages.

When you see the function name `MergeEnvFileMap` and it is passed `envFile`, it would be easy to think that the function does something with the envFile itself.

I would suggest moving the debug message for envFile into the loop, and just using `MergeMap`. When multiple `envFiles` are in use, I don't think we need to warn about values being overridden between them, as we have clearly documented the precedence... and it's intuitive when they come in a particular order on the command line.

The existing message below about `--env` overriding `--env-file` is more useful... it's less obvious which order they would apply.

dtrudg (changes requested):
Please see inline comment - I don't think we need to warn on each override across multiple env files, as the precedence is intuitive and clear.

Other than that, looks great!

internal/pkg/runtime/launcher/native/launcher_linux.go
cyanezstange: I agree, I don't feel completely comfortable with that solution. I wasn't sure if the warning was really needed and just ended being conservative about the previous code. I also agree that the documentation should be enough.
I'll work on this, thanks.

dtrudg (approved):
LGTM - Please go ahead and merge. Thanks!

PR #2915, opened 2024-05-13
singularity.conf: Add allow uts ns option

### LINKED ISSUE #1767: Disable --uts option in singularity.conf
【Request】
　In singularity.conf, add ’allow userns' and 'allow uts ns' like 'allow pid ns'.

【Background】
  EDA (Electronic design automation)  tool uses FlexLM (https://www.flexera.com/) for license management.
It is possible to restrict the access source of the license, and provides restrictions using "username" and "hostname"
 
【Theme】
  When running the EDA tool in an environment that uses singularity, singularity exec --hostname ok_machine eda.sif eda_exec
By doing so, the above restrictions can be passed by disguising the hostname.
This is an issue that should be solved by FlexLM and/or EDA tools, but the situation is that singularity cannot be introduced because the developer does not support it.
Similarly, by binding /etc/passwd, the user name is changed and the above restrictions can be passed.

cyanezstange:

`singularity.conf` now accepts setting the `allow uts ns` option, and can
invalidate the use of the `--uts` and `--hostname` flags.

 - Fixes #1767


dtrudg (changes requested):
Looks good to me - a couple of minor things that aren't related to the functionality.

dtrudg (approved):
LGTM

PR #2956, opened 2024-06-03
Add "allow ipc ns" option to singularity.conf

### LINKED ISSUE #2942: allow ipc ns in singularity.conf

After #1767 we now have both `allow pid ns` and `allow uts ns` options in `singularity.conf`.

For completeness let's add an `allow ipc ns` option, that defaults to `yes`.

When `allow ipc ns = no` then SingularityCE should not use an IPC namespace.

Singularity should warn when the IPC namespace is requested, but is not used due to the configuration option.

cyanezstange:
Add new option to singularity.conf: "allow ipc ns"

 - Fixes #2942


internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -519,6 +519,18 @@ func (e *EngineOperations) prepareContainerConfig(starterConfig *starter.Config)
 	// always set mount namespace
 	e.EngineConfig.OciConfig.AddOrReplaceLinuxNamespace(specs.MountNamespace, "")
 
+	// if IPC namespace is not allowed remove it from namespaces
+	if !e.EngineConfig.File.AllowIpcNs && e.EngineConfig.OciConfig.Linux != nil {
+		namespaces := e.EngineConfig.OciConfig.Linux.Namespaces
+		for i, ns := range namespaces {
```
dtrudg, line 525: This loop, to remove a namespace, is now repeated 3 times. It would be good, at this point, to have a function that implements the removal of a specified namespace that can be called instead of repeating the code.

dtrudg:
:+1: Code in the e2e tests looks like an elegant method to verify. Ready for review now?

dtrudg (approved):

PR #2981, opened 2024-06-10
Add "allow user ns" option to singularity.conf

### LINKED ISSUE #2941: allow user ns in singularity.conf

After #1767 we now have both `allow pid ns` and `allow uts ns` options in `singularity.conf`.

For completeness let's add an `allow user ns` option, that defaults to `yes`.

When `allow user ns = no` then SingularityCE should not start in any flow that uses a user namespace, i.e.

* With `-u / --userns`
* With `--fakeroot`
* With an unprivileged installation

Singularity must exit with a fatal error in these situations... unlike the other namespace options.

cyanezstange:
### Description of the Pull Request (PR):
Add new option to singularity.conf: "allow user ns"

Added e2e tests, making sure to fail when `allow user ns = no` in cases of any of the following flags:
- `--userns`
- `--fakeroot`

Also fails when `allow user ns = no` in `singularity.conf` and singularity was installed with the `--without-suid` option. This last one was checked manually.

 - Fixes #2941


e2e/config/config.go
```diff
@@ -237,6 +237,31 @@ func (c configTests) configGlobal(t *testing.T) {
...
+			name:           "AllowUserNsNo",
+			argv:           []string{"--userns", c.env.ImagePath, "true"},
+			profile:        e2e.UserProfile,
+			directive:      "allow user ns",
+			directiveValue: "no",
+			exit:           1,
+		},
+		{
+			name:           "AllowUserNsNoFakeroot",
+			argv:           []string{"--fakeroot", c.env.ImagePath, "true"},
```
dtrudg, line 250: Instead of `--fakeroot` as a flag, we use `e2e.FakerootProfile` in the e2e tests, by convention.

dtrudg (changes requested):
One minor request for now.

Out of interest... what _does_ happen when singularity is installed `--without-suid` and the option is enabled?

I will double check tomorrow, but if you could add that detail on the PR comment it'd be great. Thanks.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
   precedence.
 - `singularity.conf` now accepts setting new options regarding namespaces:
   - `allow ipc ns` : disable the use of the `--ipc` flag.
+  - `allow user ns` : disable the use of any `--userns` or `--fakeroot` flags,
```
dtrudg, line 13: > disable creation of user namespaces. This will prevent execution of containers with the `--userns` or `--fakeroot` flags, and unprivileged installations of SingularityCE.

e2e/config/config.go
```diff
@@ -237,6 +237,31 @@ func (c configTests) configGlobal(t *testing.T) {
 			exit:           0,
 			resultOp:       e2e.ExpectOutput(e2e.ExactMatch, "1"),
 		},
+		{
+			name:           "AllowUserNsNo",
+			argv:           []string{"--userns", c.env.ImagePath, "true"},
```
dtrudg, line 242: In e2e tests by convention we don't use `--userns`. Instead we use the `e2e.UserNamespaceProfile` which will add the flag for the test.

internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -530,6 +530,17 @@ func (e *EngineOperations) removeNamespace(namespaceType specs.LinuxNamespaceTyp
 	}
 }
 
+// isInNamespaces checks for existence of a specific namespace in the namespaces slice.
+func (e *EngineOperations) isInNamespaces(namespaceType specs.LinuxNamespaceType) bool {
```
dtrudg, line 534: `hasNamespace` is probably a bit clearer than `isInNamespace`  because this function will be called as:

`e.hasNamespace(....)`

... and the namespace is something that the engine configuration has.

dtrudg (changes requested):
A couple of minor comments. Otherwise LGTM. Thanks.

internal/pkg/runtime/engine/singularity/prepare_linux.go
```diff
@@ -530,6 +530,17 @@ func (e *EngineOperations) removeNamespace(namespaceType specs.LinuxNamespaceTyp
 	}
 }
 
+// isInNamespaces checks for existence of a specific namespace in the namespaces slice.
```
dtrudg, line 533:
```suggestion
// hasNamespaces checks for existence of a specific namespace in the namespaces slice.
```

dtrudg (approved):
LGTM. Please use the 'squash and merge' option to squash this down to a single commit when you merge.

image

PR #3031, opened 2024-06-20
fix: copy relative symlinks correctly between stages (release-4.1)

### LINKED ISSUE #2607: build: `%files` from stage fails to copy relative symlinks
**Version of Singularity**

main

**Describe the bug**

The following definition file fails to build:

```
09:38 am $ cat test.def 
BootStrap: docker
From: ubuntu:latest
Stage: devel

%post
mkdir -p /usr/mydir/mysubdir
ln -s ../bin/true /usr/mydir/true
ln -s ../../bin/true /usr/mydir/mysubdir/true

BootStrap: docker
From: ubuntu:latest
Stage: final

%files from devel
/usr/mydir
```

```
09:38 am $ sudo singularity build test.sif test.def
[sudo] password for dtrudg-sylabs: 
WARNING: 'nodev' mount option set on /tmp, it could be a source of failure during build process
INFO:    Starting build...
INFO:    Fetching OCI image...
28.2MiB / 28.2MiB [===================================================================================] 100 % 0.0 b/s 0s
INFO:    Extracting OCI image...
INFO:    Inserting Singularity configuration...
INFO:    Running post scriptlet
+ mkdir -p /usr/mydir/mysubdir
+ ln -s ../bin/true /usr/mydir/true
+ ln -s ../../bin/true /usr/mydir/mysubdir/true
INFO:    Fetching OCI image...
28.2MiB / 28.2MiB [===================================================================================] 100 % 0.0 b/s 0s
INFO:    Extracting OCI image...
INFO:    Inserting Singularity configuration...
INFO:    Copying /usr/mydir to 
FATAL:   While performing build: unable to copy files from stage to container fs: while copying [/tmp/build-temp-2697402686/rootfs/usr/mydir] to /tmp/build-temp-2218967937/rootfs/usr/mydir: invalid symlink "/tmp/build-temp-2218967937/rootfs/usr/mydir/mysubdir/true" -> "../../bin/true"
```

cyanezstange:
cherry pick #2924

In a build stage, a valid relative symlink can be created that points to a location in the rootfs that is above the location of the symlink. E.g.:

  /usr/mydir/mysubdir/true -> ../../../bin/true

This symlink is valid in the rootfs. It resolves to `/bin/true`.

If we attempt to copy `/usr/mydir` into a new build stage, then prior to this PR the symlink would cause the copy to fail.

We have been using `CopyFromTar` to copy from src `<stage1>/usr/mydir`, to `<stage2>/usr/mydir`. The `CopyFromTar` routines enforce that the target of a hard link or symlink must be under dst, or a fatal error is raised to avoid a breakout attack.

When copying between stages our boundary is not dst, but the top level of the destination stage's rootfs. We must allow links to be created that are above dst, but within the stage rootfs.

To resolve the issue, bring over enough code from
`docker/docker/pkg/archive` that we can implement a `CopyFromTarWithRoot`, that performs the same as `CopyFromTar`, except that we can specify a destination root directory. Any hard links and symlink targets under this destination root directory are valid.

Fixes #2607

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,14 @@
 ### Bug Fixes
 
 - Use ABI 3 for Apparmor profile on Ubuntu <23.10.
-
-### Bug Fixes
-
 - Avoid unnecessary copying / extraction of OCI images and Docker tarballs into
   a layout directory when they are directly accessible as a local file /
   directory.
 - Avoid unnecessary intermediate temporary image layout when building from
   Dockerfile to OCI-SIF.
+- `%files from` in a definition file will now correctly copy symlinks that point
+  to a target above the destination directory, but inside the destination stage
+  rootfs.
 
 ## 4.1.3 \[2024-05-08\]
 
--- a/LICENSE_THIRD_PARTY.md
+++ b/LICENSE_THIRD_PARTY.md
@@ -504,14 +504,6 @@ Are code from the conmon project, under the Apache License, Version 2.0.
    limitations under the License.
 ```
 
-## github.com/docker/cli
-
-The source files:
-
-* `internal/app/singularity/instance_linux.go`
-
-Contain code from the docker cli project, under the Apache License, Version 2.0.
-
 ## github.com/google/go-containerregistry
 
 The source files:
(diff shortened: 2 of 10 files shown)


cyanezstange:
Failure due to ubuntu keyserver timeout

dtrudg (approved):
LGTM - Needs to be rebased for CHANGELOG.md.

PR #2936, opened 2024-05-20
fix: avoid unneccesary temporary layout during build --oci

dtrudg:
## Description of the Pull Request (PR):

With https://github.com/sylabs/singularity/pull/2935 in place we can now translate a Docker archive tarball into an OCI-SIF without creating a redundant copy of the image as a temporary OCI layout.

We've previously been processing the buildkit tarball as an oci-archive, which still requires extraction to a temporary layout, as ggcr's tarball package doesn't accept an archived layout without a manifest.json.

Because the buildkit output tarball includes the Docker specific manifest.json, in addition to the standard OCI layout files, it can be treated as a docker-archive - now avoiding the need for the temporary
layout.

Significantly reduces disk space needed by `build --oci`, and theassociated I/O.

### This fixes or addresses the following GitHub issues:

 - Fixes #2308 

#### Before submitting a PR, make sure you have done the following:

- Read the [Guidelines for Contributing](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md), and this PR conforms to the stated requirements.
- Added changes to the [CHANGELOG](https://github.com/sylabs/singularity/blob/main/CHANGELOG.md) if necessary according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md)
- Added tests to validate this PR, linted with `make check`  and tested this PR locally with a `make test`, and `make testall` if possible (see CONTRIBUTING.md).
- Based this PR against the appropriate branch according to the [Contribution Guidelines](https://github.com/sylabs/singularity/blob/main/CONTRIBUTING.md)
- Added myself as a contributor to the [Contributors File](https://github.com/sylabs/singularity/blob/main/CONTRIBUTORS.md)

PR #3032, opened 2024-06-20
fix: avoid unneccesary temporary layout during build --oci (release-4.1)

### LINKED ISSUE #2308: remove redundant intermediate images in `build --oci` flow
Currently, building an OCI-SIF using [buildkit](https://github.com/moby/buildkit) first builds a tar file of the OCI image, and then converts that tar to an OCI-SIF by treating it as a source for `oci-archive:` URI.

We should see if we can pipe the output of buildkit straight into OCI-SIF through a pipe, to avoid the creation of the intermediate tar file.

cyanezstange:
cherry pick #2936

With #2935 in place we can now translate a Docker archive tarball into an OCI-SIF without creating a redundant copy of the image as a temporary OCI layout.

We've previously been processing the buildkit tarball as an oci-archive, which still requires extraction to a temporary layout, as ggcr's tarball package doesn't accept an archived layout without a manifest.json.

Because the buildkit output tarball includes the Docker specific manifest.json, in addition to the standard OCI layout files, it can be treated as a docker-archive - now avoiding the need for the temporary layout.

Significantly reduces disk space needed by `build --oci`, and the associated I/O.

Fixes #2308

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - Avoid unnecessary copying / extraction of OCI images and Docker tarballs into
   a layout directory when they are directly accessible as a local file /
   directory.
+- Avoid unnecessary intermediate temporary image layout when building from
+  Dockerfile to OCI-SIF.
 
 ## 4.1.3 \[2024-05-08\]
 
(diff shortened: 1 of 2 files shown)


dtrudg (approved):
e2e failures from ubuntu / canonical outage only. Will merge with failed CI.

PR #3105, opened 2024-07-11
Fix SquashFS extraction when it contains directory symlinks

### LINKED ISSUE #3084: Error while building from image: "failed to create symlink /image/rootfs/var/tmp"
**Version of Singularity**

singularity-ce version 4.1.4-jammy

**Describe the bug**

I cannot use a (certain class of) images to build from. In particular I need to encrypt existing containers for use in sensitive environments. I cannot recreate the existing container, i.e. I need to use the image file.

However the build fails with
```
FATAL:   While performing build: packer failed to pack: root filesystem extraction failed: extract command failed:
[some "WARNING:" lines]

create_inode: failed to create symlink /image/rootfs/var/tmp, because File exists
```

**To Reproduce**

Intended command is: `singularity build --pem-path=/keys/image.pub output.img input.img`, but any build fails:

- `wget -O input.img https://depot.galaxyproject.org/singularity/pigz%3A2.3.4`
- `sudo singularity build --sandbox /tmp/img input.img`

**Expected behavior**

Encrypted image (or sandbox folder) is created

**OS / Linux Distribution**

Which Linux distribution are you using? Is it up-to-date?

Linux Mint 21.3 (Ubuntu 22.04 "jammy") fully patched

**Installation Method**

DEB from GitHub: `singularity-ce_4.1.4-jammy_amd64.deb`

**Additional context**

[Log with --debug](https://github.com/user-attachments/files/16125457/debug.txt)

Console output
```
INFO:    Starting build...
FATAL:   While performing build: packer failed to pack: root filesystem extraction failed: extract command failed: WARNING: Skipping mount /etc/hosts [binds]: /etc/hosts doesn't exist in container
WARNING: Skipping mount /etc/localtime [binds]: /etc/localtime doesn't exist in container
WARNING: Skipping mount proc [kernel]: /proc doesn't exist in container
WARNING: Skipping mount /var/lib/singularity/mnt/session/tmp [tmp]: /tmp doesn't exist in container
WARNING: Skipping mount /var/lib/singularity/mnt/session/var/tmp [tmp]: /var/tmp doesn't exist in container
WARNING: Skipping mount /var/lib/singularity/mnt/session/etc/resolv.conf [files]: /etc/resolv.conf doesn't exist in container

create_inode: failed to create symlink /image/rootfs/var/tmp, because File exists
Parallel unsquashfs: Using 4 processors
391 inodes (635 blocks) to write

[==============================================================| ] 634/635  99%

created 109 files
created 54 directories
created 281 symlinks
created 0 devices
created 0 fifos
created 0 sockets
: exit status 2
```

Flamefire:

 - Fixes #3084

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -19,6 +19,7 @@ The following have contributed code and/or documentation to this repository.
 - Adam Hughes <adam@sylabs.io>, <stickmanica@gmail.com>
 - Adam Simpson <asimpson@nvidia.com>, <adambsimpson@gmail.com>
 - Afif Elghraoui <afif.elghraoui@nih.gov>
+- Alexander Grund <alexander.grund@tu-dresden.de>
 - Amanda Duffy <aduffy@lenovo.com>
 - Ana Guerrero Lopez <aguerrero@suse.com>
 - Ángel Bejarano <abejarano@ontropos.com>
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -19,7 +19,6 @@ The following have contributed code and/or documentation to this repository.
 - Adam Hughes <adam@sylabs.io>, <stickmanica@gmail.com>
 - Adam Simpson <asimpson@nvidia.com>, <adambsimpson@gmail.com>
 - Afif Elghraoui <afif.elghraoui@nih.gov>
-- Alexander Grund <alexander.grund@tu-dresden.de>
 - Amanda Duffy <aduffy@lenovo.com>
 - Ana Guerrero Lopez <aguerrero@suse.com>
 - Ángel Bejarano <abejarano@ontropos.com>
--- a/internal/pkg/build/sources/conveyorPacker_local.go
+++ b/internal/pkg/build/sources/conveyorPacker_local.go
@@ -16,11 +16,6 @@ import (
 	"github.com/sylabs/singularity/v4/pkg/sylog"
 )
 
-// LocalConveyor only needs to hold the conveyor to have the needed data to pack
-type LocalConveyor struct {
-	src string
-}
-
 // LocalPacker ...
 type LocalPacker interface {
 	Pack(context.Context) (*types.Bundle, error)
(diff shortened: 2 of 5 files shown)


dtrudg (changes requested):
Hi @Flamefire  - thanks for sharing this. However, it's not the approach we want to take.

* Relying on scanning the output of `unsquashfs` is brittle to any changes in that tool, and adds significgantly complexity to that code.
* In other code paths, where we build e.g. from an OCI source, rather than a local image file, the directories are created after image extraction, which avoids this issue. We believe we can swap the order of operations for the localimage conveyorpacker code to fix it in a much simpler manner.

I've been discussing this with @cyanezstange and he will be tackling the issue in the next week.

Flamefire:
> Relying on scanning the output of unsquashfs is brittle to any changes in that tool, and adds significant complexity to that code. > (edit... we are now using squashfs-tools-ng tools elsewhere, and may want to here at some point, so this is why parsing unsquashfs output is additionally an issue)

I made it as simple as possible with as few assumptions as possible. E.g. using `-ls` instead of `-ll` to just get a plain list of names. One issue were warnings output by the singularity sandbox run:

```
WARNING: Skipping mount /etc/hosts [binds]: /etc/hosts doesn't exist in container
WARNING: Skipping mount /etc/localtime [binds]: /etc/localtime doesn't exist in container
WARNING: Skipping mount /installed/var/singularity/mnt/session/tmp [tmp]: /tmp doesn't exist in container
WARNING: Skipping mount /installed/var/singularity/mnt/session/var/tmp [tmp]: /var/tmp doesn't exist in container
WARNING: Skipping mount /installed/var/singularity/mnt/session/etc/resolv.conf [files]: /etc/resolv.conf doesn't exist in container
```

So the only assumption I did was that the output of `unsquashfs -ls` is a list of paths, one per file, in top-down order or at least with the root path first. In case that changes (which I find rather unlikely) then directories might not get removed leading to the same behavior as before. Not too bad IMO.

> In other code paths, where we build e.g. from an OCI source, rather than a local image file, the directories are created after image extraction, which avoids this issue. We believe we can swap the order of operations for the localimage conveyorpacker code to fix it in a much simpler manner.

You are right, that is likely better

cyanezstange:
Hi @Flamefire , thank you for this PR. I have tidy some thing in #3151 which keep you as an contributor. Please let us know if you don't want to be listed in the CONTRIBUTORS.md file or have any others concerns.

Flamefire:
That's fine for me, thanks!

PR #3151, opened 2024-07-24
Fix SquashFS extraction when it contains directory symlinks

### LINKED ISSUE #3084: Error while building from image: "failed to create symlink /image/rootfs/var/tmp"

cyanezstange:

 - Fixes #3084

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@
   (`docker://`) etc.
 - Fix confusing error messages / incorrect fall-back attempt when explicit
   execution of an OCI-SIF fails.
+- Fix failing builds from local images that have symbolic links for paths that
+  are part of the base container environment (e.g. /var/tmp -> /tmp).
 
 ### Requirements
 
--- a/internal/pkg/build/sources/conveyorPacker_local_test.go
+++ b/internal/pkg/build/sources/conveyorPacker_local_test.go
@@ -1,0 +1,124 @@
+// Copyright (c) 2018-2024, Sylabs Inc. All rights reserved.
+// This software is licensed under a 3-clause BSD license. Please consult the
+// LICENSE.md file distributed with the sources of this project regarding your
+// rights to use or distribute this software.
+
+package sources_test
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/sylabs/singularity/v4/internal/pkg/build/sources"
+	"github.com/sylabs/singularity/v4/internal/pkg/image/packer"
+	"github.com/sylabs/singularity/v4/internal/pkg/util/fs"
+	"github.com/sylabs/singularity/v4/pkg/build/types"
+)
+
+func createArchiveFromDir(dir string, t *testing.T) *os.File {
+	mk, err := exec.LookPath("mksquashfs")
+	if err != nil {
+		t.SkipNow()
+	}
+	f, err := os.CreateTemp("", "archive-")
+	if err != nil {
+		t.Fatal(err)
+	}
+	cmd := exec.Command(mk, dir, f.Name(), "-noappend", "-no-progress")
+	if err := cmd.Run(); err != nil {
+		t.Fatal(err)
+	}
+	return f
+}
+
+func makeDir(path string, t *testing.T) {
+	if err := os.MkdirAll(path, 0o755); err != nil {
+		t.Fatalf("while creating directory %s: %v", path, err)
+	}
+}
+
+func isExist(path string) bool {
+	result, _ := fs.PathExists(path)
+	return result
+}
+
+func TestSquashfsInput(t *testing.T) {
+	if s := packer.NewSquashfs(); !s.HasMksquashfs() {
+		t.Skip("mksquashfs not found, skipping")
+	}
+
+	dir, err := os.MkdirTemp(os.TempDir(), "test-localpacker-squashfs-")
+	if err != nil {
+		t.Fatalf("while creating tmpdir: %v", err)
+	}
+	defer os.RemoveAll(dir)
+
+	// Check that folder symlinks in the image to common folders (in base env) work
+	inputDir := filepath.Join(dir, "input")
+
+	makeDir(filepath.Join(dir, "var", "tmp"), t)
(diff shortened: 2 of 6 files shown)


internal/pkg/build/sources/conveyorPacker_local_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"testing"
+
+	"github.com/sylabs/singularity/v4/internal/pkg/build/sources"
+	"github.com/sylabs/singularity/v4/internal/pkg/image/packer"
+	"github.com/sylabs/singularity/v4/internal/pkg/util/fs"
+	"github.com/sylabs/singularity/v4/pkg/build/types"
+)
+
+func createArchiveFromDir(dir string, t *testing.T) *os.File {
+	mk, err := exec.LookPath("mksquashfs")
```
dtrudg, line 22: require.Command(t, "mksquashfs")

CHANGELOG.md
```diff
@@ -40,6 +40,8 @@
   (`docker://`) etc.
 - Fix confusing error messages / incorrect fall-back attempt when explicit
   execution of an OCI-SIF fails.
+- Fix failing build process when building from an image that has symbolic
```
dtrudg, line 43: This needs to be a bit more specific. Something like...

> Fix failing builds from local images that have symbolic links for paths that are part of the base container environment (e.g. `/var/tmp` -> `/tmp`).

dtrudg (approved):
Looks good to me, with a minor change required to `CHANGELOG.md` please.

Let's discuss how to squash it... and work through that together.

internal/pkg/build/sources/conveyorPacker_local_test.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if err := os.Mkdir(filepath.Join(rootfs, "var"), 0o755); err != nil {
+		t.Fatalf("while creating directory: %v", err)
+	}
+
+	// Create symlinks: /var/tmp -> /tmp , /var/log -> /tmp
+	if err := os.Symlink(filepath.Join(rootfs, "tmp"), filepath.Join(rootfs, "var", "tmp")); err != nil {
+		t.Fatalf("while creating symlink: %v", err)
+	}
+	if err := os.Symlink(filepath.Join(rootfs, "tmp"), filepath.Join(rootfs, "var", "log")); err != nil {
```
Flamefire, line 48: `rootfs` is an absolute path to somewhere on the temporary folder, isn't it? I'd suggest to use the original one with `../tmp` to match the original issue or `/tmp` directly to match the comment.
```suggestion
	if err := os.Symlink(filepath.Join("..", "tmp"), filepath.Join(rootfs, "var", "tmp")); err != nil {
		t.Fatalf("while creating symlink: %v", err)
	}
	if err := os.Symlink(filepath.Join("..", "tmp"), filepath.Join(rootfs, "var", "log")); err != nil {
```

cyanezstange: Hi, thank you very much for the review and pointing out this issue. I'll check it soon.

cyanezstange:
Changed targets for unit test symlinks as suggested by @Flamefire . Links now point to a proper directory as expected.
There is an e2e test that fails due to a Docker rate-limit, what's your suggestion on this @dtrudg ?

Flamefire (approved):

dtrudg:
> Changed targets for unit test symlinks as suggested by @Flamefire . Links now point to a proper directory as expected. There is an e2e test that fails due to a Docker rate-limit, what's your suggestion on this @dtrudg ?

So - really it doesn't matter where the symlink points. The issue is just caused by a symlink being present. Using `<rootfs>/tmp` does give a 'broken' symlink in the container, but this will still cause the issue to present itself.

It's fine to use the relative symlink, or absolute symlink though. We aren't doing anything here which is problematic.

I've re-triggered the failing test. If it passes then this is good to merge.

PR #3160, opened 2024-07-25
chore: refactor mksquashfs code (release-4.1)

cyanezstange:

Backport #3122

Code calling out to mksquashfs was previously tightly integrated with image building. To allow mksquashfs to be called elsewhere, such as from the ongoing OCI-SIF overlay work, refactor it out to pkg/util/fs/squashfs.

Note - checking for `-comp` flag support is removed, as mksquashfs is now new enough in our target distributions. We just default to requesting `-comp gzip`.


dtrudg (approved):
Looks good to me, assuming CI pass.

PR #3165, opened 2024-07-26
Fix SquashFS extraction when it contains directory symlinks (release 4.1)

cyanezstange:
Backport #3151 


dtrudg (approved):
LGTM pending confirmation it's okay to add Alexander Grund to CONTRIBUTORS.md.

PR #3267, opened 2024-08-27
fix: hack to output OCI v1.1 artifact

tri-adam:

Data containers in Singularity use an empty JSON document as their "config". Thus, based on the [OCI v1.1 release notes](https://opencontainers.org/posts/blog/2024-03-13-image-and-distribution-1-1/), they're a good candidate to make use of the `artifactType` field.

The `google/go-containerregistry` does not currently include `artifactType` in `v1.Manifest`, as per the OCI v1.1 specification. There is an open PR to address that (https://github.com/google/go-containerregistry/pull/1931) but it hasn't had any movement in a while. This PR works around the lack of support until that field is added in the `google/go-containerregistry` module.

 - Fixes #


internal/pkg/ocisif/datacontainer.go
```diff
@@ -158,3 +167,68 @@ func DataContainerLayerOffset(f *os.File) (int64, error) {
...
+
+	var manifest struct {
+		SchemaVersion int64             `json:"schemaVersion"`
+		MediaType     types.MediaType   `json:"mediaType,omitempty"`
+		ArtifactType  string            `json:"artifactType,omitempty"`
+		Config        v1.Descriptor     `json:"config"`
+		Layers        []v1.Descriptor   `json:"layers"`
+		Annotations   map[string]string `json:"annotations,omitempty"`
+		Subject       *v1.Descriptor    `json:"subject,omitempty"`
+	}
```
wobito, line 215: Can't simplify this a bit with?

```suggestion
	var manifest struct {
		v1.Manifest
		ArtifactType  string            `json:"artifactType,omitempty"`
	}
```

wobito (commented):
just noticed one thing

internal/pkg/ocisif/datacontainer.go
tri-adam: That's definitely simpler, but I think that will change the order of the fields? That doesn't matter at the JSON level, but that would mean that we'd generate a different manifest digest than GGCR likely will (if/when https://github.com/google/go-containerregistry/pull/1931 lands.)

wobito: That makes sense

dtrudg (approved):
LGTM. Many thanks for this!

tri-adam:
Just had one more thought this morning, we are setting the `artifactType` to `application/vnd.sylabs.data-container.config.v1+json`... but the artifact itself is not a `config`, nor is it JSON document? The layer media types describe the format of the layer, so I don't think we need to describe that here, but perhaps this should just be something like `application/vnd.sylabs.data-container.v1`?

dtrudg:
> Just had one more thought this morning, we are setting the `artifactType` to `application/vnd.sylabs.data-container.config.v1+json`... but the artifact itself is not a `config`, nor is it JSON document? The layer media types describe the format of the layer, so I don't think we need to describe that here, but perhaps this should just be something like `application/vnd.sylabs.data-container.v1`?

Ahh, yeah. That would make sense without the `config` 👍

PR #3502, opened 2025-01-28
feat: libsubid support (revised)

### LINKED ISSUE #3434: Pick libsubid support from Apptainer to SingularityCE
Pick https://github.com/apptainer/apptainer/pull/2433 from apptainer with any required adapation for native mode in SingularityCE.

### LINKED ISSUE #3435: Adapt libsubid support for OCI mode in SingularityCE
Continue from #3434 to add libsubid support in OCI-mode (not present in Apptainer).

dtrudg:

Pick and adapt https://github.com/apptainer/apptainer/pull/2433 with heavy revision.

Original description:

> This commit adds support for remote access to /etc/subuid and
> /etc/subgid mappings from ldap services using libsubid. Requires
> shadow-utils + libsubid.
>   
> Code ported from podman / container storage.

Adapted for SingularityCE:

- When compiled with libsubid support (default), subid / gid ranges used in `--fakeroot` and `--oci` mode etc. will *only* be retrieved via libsubid. This replicates the behaviour of the shadow subid utilities. If `nsswitch.conf` is set to a non-files subid source then there is no fallback to the local `/etc/sub[ug]id` files.
- If compiled without libsubid support (mconfig `--without-libsubid` flag)  the previous behaviour of consulting `/etc/sub[ug]id` files, regardless of system `nsswitch.conf` configuration, remains.
- The fakeroot plugin callback for providing a custom ID mapping is removed. This allows code to be simplified, and using libsubid now provides a route to non-local-file id mappings.
- A mutex on libsubid calls is added to avoid test failures as it is non-thread safe.

Note that the `config fakeroot` command will still always operate directly on local `/etc/sub[ug]id` files. There is no support for modifying configured mappings through libsubid.

**Build & Packaging Implications**

* Makes libsubid support the default, and fail if libsubid headers are not available.
* Adds a `--without-libsubid` flag to `mconfig` to build without libsubid support.

Adopting libsubid support as a required default ensures that, unless built with `--without-libsubid`, SingularityCE will honor any nss subid configuration on a system.

* Builds EL RPM packages with libsubid support.
* Builds Ubuntu Deb packages without libsubid support.
* Drops SLES reated rules from RPM spec.

Our target EL distributions all offer libsubid support, so ensure RPM packages are built with it.

Only Ubuntu 24.04 offers libsubid support - so build Deb packages without it.

The INSTALL.md hasn't included 'SLES / openSUSE' as a supported target of the RPM spec for a long time. We also don't test the rules there, and they are suspected broken.

 - Fixes #3434 
 - Fixes #3435 


tri-adam (approved):
LGTM

PR #3533, opened 2025-02-14
chore: Use go 1.24.0 in CI, bump golangci-lint, fix lint

dtrudg:
Bump Go version used in CI to latest 1.24.0

Clarify requirement for >=1.23.4

Bump golangci-lint and fix much newly reported lint.

Most of the lint is related to the Go `os.XXX` routines giving `int` uid/gid/modes, while lower level calls, 3rd party things we call, and structs we use, use `uint32` - which is actually what uid/gid are on Linux. Singularity code has a mix of using `int` and `uint32` to bridge between the worlds.

Unfortunately dependencies are pulling in a patch version above 1.23.0 at this time.


.circleci/config.yml
```diff
@@ -41,7 +41,7 @@ executors:
       - image: ubuntu:24.04
   golangci-lint:
     docker:
-      - image: golangci/golangci-lint:v1.60.1
+      - image: golangci/golangci-lint:v1.64.5
```
tri-adam, line 44: Should the patch version be pinned here? Might be better to always pull in the latest patch for a given major/minor?
```suggestion
      - image: golangci/golangci-lint:v1.64
```

tri-adam (approved):

PR #3641, opened 2025-04-23
chore: switch to golangci-lint v2.1

dtrudg:
Bump the linter, and fix new lint.

Note - for review, please especially scrutinize the negated conditional changes, and the if/else -> switch changes to satisfy the effective flow is unchanged.


tri-adam (approved):
LGTM

PR #4095, opened 2026-04-29
Configurable root / non-root search paths for external binaries

dtrudg:

Prior to this PR, except for executables that have their own explicit path entry in singularity.conf, we search the value of $PATH in the environment, plus sensible default directories.

Although singularity is not intended to contain / prevent host modification when run as host-root, there is a valid defence-in-depth argument for performing / allowing search path sanitization.

Add `user search path` and `root search path` to `singularity.conf`, which control where bin.FindOnPath looks for things.

When started as host root, external binaries (except those with explicit configuration entries) are now found using the `root search path` in `singularity.conf`. By default this excludes searching the environment `$PATH`. Add `$PATH:` to the start of `root search path` in `singularity.conf` to restore previous behavior.

When started as non-root / fake root, external binaries (except those with explicit configuration entires) are now found using the `user search path` in `singularity.conf`. By default this includes `$PATH`, so there is no effective behaviour change vs previous versions.

Thank you to @KoseceMehmet for suggesting this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # SingularityCE Changelog
 
+## Unreleased
+
+## Change Defaults / Behaviours
+
+Although SingularityCE does not aim to contain execution / prevent host
+modification when started as the host root user, the following changes have been
+adopted to permit finer control over the use of external binaries, with a
+modified default search path when `singularity` is run as the host root user:
+
+- When started as host root, external binaries (except those with explicit
+  configuration entries) are now found using the `root search path` in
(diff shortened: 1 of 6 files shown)


KoseceMehmet:
Hi @dtrudg,
Thank you for implementing this change and for the mention. I'm glad to see that the defense-in-depth argument regarding search path sanitization led to a more secure configuration model for SingularityCE. I believe this significantly strengthens the trust boundary for privileged executions.

Best regards

josueneo (approved):

PR #4162, opened 2026-06-08
chore: tidy up layout.Manager code

dtrudg:

Tidy up the layout.Manager code a bit, before we do further work on it.

Drop the VFS interface. We've only ever had, and will only have one VFS implementation.

Move the VFS implementation into a separate vfs.go file.

Create a Manager via NewManager, rather than bare struct. Set the root path in NewManager.

In TestLayout, the session var is actually a Manager... so call it mgr to be less confusing!

### DIFF
--- a/internal/pkg/util/fs/layout/manager.go
+++ b/internal/pkg/util/fs/layout/manager.go
@@ -7,12 +7,10 @@ package layout
 
 import (
 	"fmt"
-	iofs "io/fs"
 	"os"
 	"path/filepath"
 	"slices"
 	"strings"
-	"syscall"
 
 	"github.com/sylabs/singularity/v4/internal/pkg/util/fs"
 )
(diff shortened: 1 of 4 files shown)


internal/pkg/util/fs/layout/manager.go
josueneo, line 124: Since `setRootPath` is only called by `NewManager` and is unexported, then `m.entries` is always nil at this point. That else block is unreachable, right?

Same case for `m.ovDirs` and `m.dirs`?

internal/pkg/util/fs/layout/manager.go
```diff
@@ -177,8 +112,8 @@ func (m *Manager) createParentDir(path string) {
 	}
 }
 
-// SetRootPath sets layout root path
-func (m *Manager) SetRootPath(path string) error {
+// setRootPath sets layout root path
+func (m *Manager) setRootPath(path string) error {
 	if !fs.IsDir(path) {
 		return fmt.Errorf("%s is not a directory or doesn't exists", path)
```
josueneo, line 118: Small, minor typo? `exist` vs `exists` worth changing while the code is touched?

josueneo (approved):

internal/pkg/util/fs/layout/manager.go
dtrudg: So this is an initial coarse tidy up trying to minimise code change. There will be refactoring follow ups which change the initialisation flow etc.

internal/pkg/util/fs/layout/manager.go
dtrudg: Yeah this is a typo in the existing code prior to this pr. I can change it here, but setRootPath will disappear in further refactoring.

PR #4164, opened 2026-06-11
rpc: use filepath.rel / isLocal instead of string manipulations

dtrudg:

Replace string manipulation in Mount / Mkdir RPCs with rel/Islocal.


dtrudg:
Closing - overlooked the flow where paths are absolute w.r.t. the session dir.

